PR #112, opened 2016-10-18
Fix lastModified

pedrohfernandes:
The method StatObject not returns the last modified parameter.  The lastModified variable was not being set.


harshavardhana:
Thanks @pedrohfernandes

PR #117, opened 2017-02-17
API works with Netcore

poornas:
Minio API implementation for .NetCore in MinioCore2 xproj.  This still needs to be unified to use single code base for .NET4.5.2 and .NetCore


harshavardhana:
Can you fix all the conflicts here?

Minio/AWSS3Endpoints.cs
```diff
@@ -0,0 +1,75 @@
+﻿using System;
```
harshavardhana: Missing license header.

poornas: Both Core and Framework run under Minio namespace. The issue here is that code in MinioCore2 is a .NetCore class library that compiles for both Core and Framework, however the Core tests run fine while the Framework give AWS signature mismatch ( this is after stripping the port from host name, that was tripping me up earlier).

Maybe the better strategy would be to merge to minio-dotnet after Core and Framework code base is unified. Nitish can work from my fork until then.

harshavardhana:
Closing the PR, created a new branch `netcore`

Wahid101:
Does the `netcore ` branch still exist? I couldn't find it

poornas:
It was merged into master branch - if you are looking for the nuget package, both .NETCore and .NET framework would be in https://www.nuget.org/packages/Minio/

Wahid101:
@poornas maybe you can help me out. Im trying to get the latest .net core lib, but nugget shows the latest (1.1.1) published April of last year while the non core was released (3.0.2) last week. I want to make sure I'm not missing any important updates so I cloned this project, but I don't see anything related to .net core

poornas:
the package Minio on nuget will work for .NetCore projects as well @Wahid101

Wahid101:
just so I understand, I should use the package named `Minio 3.0.2` and not `Minio.NetCore 1.1.1` in my .net core 2.1 project?

poornas:
that's correct

PR #132, opened 2017-05-05
Nuspec for .NetCore and .NET4.5.2 as separate packages

poornas:
Separate Nuget packages for the Minio SDK - one for .Netcore, the other for .NET4.5.2.  This is required because of inconsistencies in Nuget tool causing package errors when the SDK is imported into .Netcore project.

### DIFF
--- a/Minio.Core/Minio.Core.csproj
+++ b/Minio.Core/Minio.Core.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp1.0</TargetFramework>
+    <TargetFramework>netstandard1.6</TargetFramework>
     <AssemblyName>Minio</AssemblyName>
     <GeneratePackageOnBuild>False</GeneratePackageOnBuild>
     <RootNamespace>Minio</RootNamespace>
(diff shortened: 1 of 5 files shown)


harshavardhana:
The changes look good, going to test this on windows. @poornas

deekoder:
@poornas - I am not the right person to review this PR. I will reassign to someone who is more knowledgeable in this area.

donatello:
Some questions: 

1. Should there be separate documentation on how to use each of the separate packages?
2. Also, I would appreciate steps to test it out each of the different packages.

harshavardhana:
> Does the binary .nuget/nuget.exe need to be checked into the repo? It is 4+ MB.

No we shouldn't @donatello

poornas:
It was left in because latest nuget version recommended on their website doesnt work for VS2017, and every time some one builds the package this confusion could arise until the recommended version catches up. I can remove the folder, however we need to have a way of documenting dev. dependencies.

harshavardhana:
> It was left in because latest nuget version recommended on their website doesnt work for VS2017, and every time some one builds the package this confusion could arise until the recommended version catches up. I can remove the folder, however we need to have a way of documenting dev. dependencies.

We should document it precisely, we shouldn't keep binaries in git repo.

poornas:
@donatello , there is nothing different about the usage of the packages. There are just meant to be used by .NET4.5.2 apps or .NetCore apps.  On VS 2017, you can create console apps targeting the .NetCore framework or .Net4.5.2 framework, and import the corresponding Minio-dotnet SDK  for testing.

harshavardhana:
ping @donatello any feedback?

PR #134, opened 2017-05-09
Disable SHA-256 computation for https

### LINKED ISSUE #101: Disable sha256 computation for https

poornas:
Currently minio-dotnet sdk computes sha256 for https requests. This change fixes #101.

### DIFF
--- a/Minio/V4Authenticator.cs
+++ b/Minio/V4Authenticator.cs
@@ -30,7 +30,7 @@ namespace Minio
     {
         private readonly string accessKey;
         private readonly string secretKey;
-
+        private string scheme;
         //
         // Excerpts from @lsegal - https://github.com/aws/aws-sdk-js/issues/659#issuecomment-120477258
         //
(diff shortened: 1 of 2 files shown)


Minio/V4Authenticator.cs
```diff
@@ -452,8 +452,15 @@ private void SetHostHeader(IRestRequest request, IRestClient client)
         /// Set 'x-amz-content-sha256' http header.
         /// </summary>
         /// <param name="request">Instantiated request object</param>
-        private void SetContentSha256(IRestRequest request)
+        private void SetContentSha256(IRestRequest request, IRestClient client)
         {
+            // No need to compute SHA256 if endpoint scheme is https
+            if (client.BaseUrl.Host.StartsWith("https"))
```
harshavardhana: We should have the client struct remember if the RestClient is secure or not. It is odd style to passing the entire IRestClient into 'SetContentSha256' and unsafe - as a thumb rule we should only passing the necessary arguments.

PR #139, opened 2017-05-27
Fix: Avoid md5 hash calculation for signed payloads

### LINKED ISSUE #130: For all signed payloads with sha256 avoid calculating Content-Md5

poornas:
Fixes #130 
For authenticated requests that are secure set sha-256 hash to "UNSIGNED-PAYLOAD" and
calculate md5 hash as well. Authenticated but insecure requests use strict sha-256 hash.

Anonymous requests both secure and insecure need to calculate md5-hash.


harshavardhana:
For multi object deletes `Content-Md5` should be set in accordance with S3 protocol can you validate this with our SDK ? this is regardless of signed payloads with sha256.

harshavardhana:
> For multi object deletes Content-Md5 should be set in accordance with S3 protocol can you validate this with our SDK ? this is regardless of signed payloads with sha256.

I see that RemoveObjects() is not implemented yet.

poornas:
@y4m4b4, RemoveObjects is not implemented as yet for dotnet sdk

harshavardhana:
@donatello pending your review..

Minio/V4Authenticator.cs
```diff
@@ -463,11 +460,13 @@ private void SetHostHeader(IRestRequest request, string hostUrl)
...
         {
             // No need to compute SHA256 if endpoint scheme is https
-            if (this.scheme.Equals("https"))
+            if (isSecure && !isAnonymous)
             {
                 request.AddHeader("x-amz-content-sha256","UNSIGNED-PAYLOAD");
                 return;
             }
+            if (this.isAnonymous)
+                return;
```
donatello: You can simplify the code by reordering a bit: put the above if block as the first one in the function, then adding the unsigned-payload header if condition may just check if isSecure is true.

Minio/V4Authenticator.cs
```diff
@@ -514,6 +513,10 @@ private void SetContentMd5(IRestRequest request)
                 {
                     return;
                 }
+                // For insecure, authenticated requests calculate sha256 only
```
donatello: Commenting just about sha256 in the md5 header function is out of place. It should say something like "MD5 is not required if in insecure and authenticated mode, since the sha256 header would be computed".

poornas: updated PR to apply review comments

Minio/V4Authenticator.cs
```diff
@@ -69,22 +70,18 @@ internal class V4Authenticator : IAuthenticator
         /// <summary>
         /// Authenticator constructor.
         /// </summary>
+        /// <param name="scheme">URL Scheme</param>
         /// <param name="accessKey">Access key id</param>
         /// <param name="secretKey">Secret access key</param>
-        public V4Authenticator(string accessKey, string secretKey)
+        public V4Authenticator(string scheme,string accessKey, string secretKey)
         {
+            this.isSecure = (scheme == "https") ? true : false;
```
donatello: It is better for scheme to directly be a bool instead of string here.

Issue #142, opened 2017-06-16
Feature: Add put object with meta data Api


poornas:
Catch up with minio-go sdk

PR #143, opened 2017-06-19
Support metadata for PutObject and StatObject calls

poornas:
Allows custom metadata to be set and retrieved via Put and Stat calls. This is w.r.t #142

### DIFF
--- a/Docs/API.md
+++ b/Docs/API.md
@@ -765,7 +765,7 @@ catch (MinioException e)
 <a name="putObject"></a>
 ### PutObjectAsync(string bucketName, string objectName, Stream data, long size, string contentType)
 
-` Task PutObjectAsync(string bucketName, string objectName, Stream data, long size, string contentType, CancellationToken cancellationToken = default(CancellationToken))`
+` Task PutObjectAsync(string bucketName, string objectName, Stream data, long size, string contentType,Dictionary<string,string> metaData=null, CancellationToken cancellationToken = default(CancellationToken))`
 
 
 Uploads contents from a stream to objectName.
@@ -781,6 +781,8 @@ __Parameters__
 | ``data``  | _Stream_  | Stream to upload |
 | ``size``  | _long_    | size of stream   |
 | ``contentType``  | _string_ | Content type of the file. Defaults to "application/octet-stream" |
+| ``metaData``  | _Dictionary<string,string>_ | Dictionary of metadata headers. Defaults to null. |
+
 | ``cancellationToken``| _System.Threading.CancellationToken_ | Optional parameter. Defaults to default(CancellationToken) |
 
 
@@ -822,7 +824,7 @@ catch(MinioException e)
 <a name="putObject"></a>
 ### PutObjectAsync(string bucketName, string objectName, string filePath, string contentType=null)
 
-` Task PutObjectAsync(string bucketName, string objectName, string filePath, string contentType=null, CancellationToken cancellationToken = default(CancellationToken))`
+` Task PutObjectAsync(string bucketName, string objectName, string filePath, string contentType=null,Dictionary<string,string> metaData=null, CancellationToken cancellationToken = default(CancellationToken))`
 
 
 Uploads contents from a file to objectName.
@@ -837,6 +839,8 @@ __Parameters__
 | ``objectName``  | _string_  | Object name in the bucket |
 | ``fileName``  | _string_  | File to upload |
 | ``contentType``  | _string_ | Content type of the file. Defaults to " |
+| ``metadata``  | _Dictionary<string,string>_ | Dictionary of meta data headers and their values.Defaults to null.|
+
 | ``cancellationToken``| _System.Threading.CancellationToken_ | Optional parameter. Defaults to default(CancellationToken) |
 
 
--- a/Minio.Functional.Tests/FunctionalTest.cs
+++ b/Minio.Functional.Tests/FunctionalTest.cs
@@ -379,7 +379,9 @@ namespace Minio.Functional.Tests
             ObjectStat statObject = await PutObject_Tester(minio, bucketName, objectName, fileName, contentType:contentType, metaData:metaData);
             Assert.IsTrue(statObject != null);
             Assert.IsTrue(statObject.metaData != null);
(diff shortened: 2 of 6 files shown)


Minio/MinioClient.cs
```diff
@@ -206,14 +206,14 @@ private string FullUserAgent
 
             }
 
-            if (contentType != null)
-            {
-                request.AddHeader("Content-Type", contentType);
-            }
+            //if ((contentType != null) && (headerMap == null || !headerMap.ContainsKey("Content-Type")))
```
harshavardhana: Is this intentional?

Docs/API.md
```diff
@@ -781,6 +781,8 @@ __Parameters__
 | ``data``  | _Stream_  | Stream to upload |
 | ``size``  | _long_    | size of stream   |
 | ``contentType``  | _string_ | Content type of the file. Defaults to "application/octet-stream" |
+| ``metaData``  | _Dictionary<string,string>_ | Dictionary of metadata headers. Defaults to null. |
```
krisis: The description for `metadata` could be,

Object metadata to be stored. Defaults to _null_.

Docs/API.md
```diff
@@ -837,6 +839,8 @@ __Parameters__
 | ``objectName``  | _string_  | Object name in the bucket |
 | ``fileName``  | _string_  | File to upload |
 | ``contentType``  | _string_ | Content type of the file. Defaults to " |
+| ``metadata``  | _Dictionary<string,string>_ | Dictionary of meta data headers and their values.Defaults to null.|
```
krisis: The description for `metadata` could be,

Object metadata to be stored. Defaults to _null_.

Minio/ApiEndpoints/IObjectOperations.cs
```diff
@@ -53,9 +53,10 @@ public interface IObjectOperations
         /// <param name="data">Stream of file to upload</param>
         /// <param name="size">Size of stream</param>
         /// <param name="contentType">Content type of the new object, null defaults to "application/octet-stream"</param>
+        /// <param name="metaData">Dictionary of metadata headers. Optional, defaults to null</param>
         /// <param name="cancellationToken">Optional cancellation token to cancel the operation</param>
 
-        Task PutObjectAsync(string bucketName, string objectName, Stream data, long size, string contentType = null, CancellationToken cancellationToken = default(CancellationToken));
+        Task PutObjectAsync(string bucketName, string objectName, Stream data, long size, string contentType = null, Dictionary<string, string> metadata=null, CancellationToken cancellationToken = default(CancellationToken));
```
krisis: This will break applications that were written based on our previous API. Can we salvage this by using optional arguments feature in C#? E.g,

```cs
Task PutObjectAsync(string bucketName, string objectName, Stream data, long size, string contentType = null, CancellationToken cancellationToken = default(CancellationToken), Dictionary<string, string> metadata=null);
```
Ref: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/named-and-optional-arguments#optional-arguments

Minio/MinioClient.cs
poornas: unintentional - removing from commit.

Docs/API.md
poornas: @krisis, fixed..

Minio.Functional.Tests/FunctionalTest.cs
```diff
@@ -364,8 +364,31 @@ public static void Main(string[] args)
...
+            string fileName = CreateFile(1 * MB);
+            string contentType = "custom/contenttype";
+            Dictionary<string, string> metaData = new Dictionary<string, string>(){
+                { "x-amz-meta-customheader", "minio-dotnet"}
+            };
+            await Setup_Test(minio, bucketName);
+            ObjectStat statObject = await PutObject_Tester(minio, bucketName, objectName, fileName, contentType:contentType, metaData:metaData);
+            Assert.IsTrue(statObject != null);
+            Assert.IsTrue(statObject.metaData != null);
+            Assert.IsTrue(statObject.metaData.ContainsKey("x-amz-meta-customheader"));
```
krisis: `Assert.IsTrue(statObject.metaData.ContainsKey("x-amz-meta-customheader"));`

The above assertion fails. The fix is,

`Assert.IsTrue(statObject.metaData.ContainsKey("X-Amz-Meta-Customheader"));`

Minio/ApiEndpoints/IObjectOperations.cs
```diff
@@ -54,8 +54,8 @@ public interface IObjectOperations
         /// <param name="size">Size of stream</param>
         /// <param name="contentType">Content type of the new object, null defaults to "application/octet-stream"</param>
         /// <param name="cancellationToken">Optional cancellation token to cancel the operation</param>
-
-        Task PutObjectAsync(string bucketName, string objectName, Stream data, long size, string contentType = null, CancellationToken cancellationToken = default(CancellationToken));
+        /// <param name="metaData">Object metadata to be stored. Defaults to null.</param>
```
krisis: parameter description should indicate that `metaData` is also an optional parameter.

Minio.Functional.Tests/FunctionalTest.cs
poornas: @krisis, this fails for me if I camelcase it.  Did you just run the functional test against play server without changes? When the header is lowercased in the PUT call, it should get saved as is and be returned by Stat call.

krisis: @poornas, it fails when run as is for me. I can test again and share my output. Would you need anything else to debug?

poornas: @krisis, can you run the functional tests again? I was using case sensitive comparison of the custom metadata headers though it is not guaranteed by http header response. I fixed the functional test as there is no real issue with the statobject/pubobject calls per se.

krisis: @poornas OK, I shall run the functional test and confirm. I wonder why you weren't able to see the assertion failure in your development setup. Do you know if header key comparison behaviour changed between dotnet versions?

PR #146, opened 2017-06-26
Fix: Reference nuget package instead of local copy of project

poornas:
The functional test project needs to reference published version of minio.netcore instead of referencing local copy of project. This will help in dockerizing the functional tests.


krisis:
@poornas how will developers run functional tests on local code changes to Minio dot-net SDK that are not in nuget (yet)?

poornas:
@krisis, agree - it is already an issue  because appveyor failure here is due to nuget version not having local code changes. We might have to keep the release version used by mint as a separate project and one for developer testing - however code maintainability questions arise.

krisis:
>@krisis, agree - it is already an issue because appveyor failure here is due to nuget version not having local code changes. We might have to keep the release version used by mint as a separate project and one for developer testing - however code maintainability questions arise.

@poornas, how are we going forward with this PR?

PR #154, opened 2017-07-20
Change order of cancellation token and metadata in sync with Api doc

### LINKED ISSUE #153: Support for metadata not in .netCore nuget package
I discovered that the .netCore Nuget package contains no PutObjectAsync method with metadata in ObjectOperations.cs

To fix this issue I checked out the git repo and used the sources directly. However there the method signature is:
```
public async Task PutObjectAsync(string bucketName, string objectName, string fileName, string contentType = null, CancellationToken cancellationToken = default(CancellationToken), Dictionary<string, string> metaData = null) 
```

In the [API.md](https://github.com/minio/minio-dotnet/blob/master/Docs/API.md#putobjectasyncstring-bucketname-string-objectname-stream-data-long-size-string-contenttype) however the metaData argument comes before the CancellationToken.

poornas:
Fixes #153. The API documentation correctly gives cancellation token as last argument to the PutObjectAsync call, whereas the implementation put metaData as the last argument. Change the order of arguments in the implementation to be consistent with documentation. There is no impact to users because this feature has not been published to Nuget yet.

### DIFF
--- a/Minio.Functional.Tests/FunctionalTest.cs
+++ b/Minio.Functional.Tests/FunctionalTest.cs
@@ -101,6 +101,7 @@ namespace Minio.Functional.Tests
 
                 // Set HTTP Tracing Off
                 // minioClient.SetTraceOff();
+                PutObject_Test5(minioClient).Wait();
 
                 // Check if bucket exists
                 BucketExists_Test(minioClient).Wait();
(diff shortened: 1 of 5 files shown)


poornas:
@donatello, waiting for your review on this.

PR #166, opened 2017-08-25
Update Functional tests to new Mint log format

### LINKED ISSUE #160: Update minio-dotnet to output new mint log format
`minio-dotnet` tests are used in `mint` to test Minio server. `minio-dotnet` logs need to be updated in the format below so that `mint` logs can be easily parsed.
```
{
    "name":"mc", // SDK Name
    "function":"testPresignGetObject", // Test function name
    "description": "test function description (optional)", //  Test function description
    "args":"", // key value map, varName:value. Only arguements that devs may be interested in
    "duration":"", // duration of the whole test
    "status":"", // can be PASS, FAIL, NA
    "alert":"failed to download pre-signed object(optional)", // error related, human readable message. Should be taken care of if present
    "error":"stack-trace/exception message(only in case of failure)" // actual low level exception/error thrown by the program
}
```

poornas:
Fixes #160 

Retaining "description" as a field for the logger as it is useful to make out what is being tested. It is not written out to conform with the latest discussion on the log format in https://github.com/minio/mint/issues/114.

### DIFF
--- a/Minio.Functional.Tests/FunctionalTest.cs
+++ b/Minio.Functional.Tests/FunctionalTest.cs
@@ -109,132 +109,150 @@ namespace Minio.Functional.Tests
             else
                 minioClient = new MinioClient(endPoint, accessKey, secretKey);
 
-            try
+
+            // Assign parameters before starting the test 
+            string bucketName = GetRandomName();
+            string objectName = GetRandomName();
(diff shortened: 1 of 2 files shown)


deekoder:
@poornas travis fails.

poornas:
@deekoder , this needs reviewers.

PR #165, opened 2017-08-23
Fix: use default region if bucket not in cache

poornas:
Fixes #164 

Region was not being initialized correctly for presigned post policy operation.

PR #167, opened 2017-08-25
Fix adding dotnet repo in mono_install script

vadmeste:


harshavardhana:
BTW @poornas why is appveyor failing here?

poornas:
@harshavardhana , dotnet CLI now enforces package version - System.Net.Http uses 4.3.1 version of package whereas 4.3.2 is available. It used to raise a warning flag, but now fails instead. The package version on the Minio.Client.Examples.csproj needs to be upped to 4.3.2. https://github.com/minio/minio-dotnet/pull/165 passes Appveyor because package version was fixed on the project.

PR #173, opened 2017-09-18
Fix PutObject call for unknown size stream

### LINKED ISSUE #169: PutObject call needs to support unknown sized stream
SDK needs to catch up on unknown sized stream support.

Needs functional test - see #156 
 3. Missing test for putobject() for unknown sized stream (single/multipart) (may be missing feature)

poornas:
Fixes #169 

Currently PutObject call with unknown stream size loops infinitely. Fixed that bug and added a functional test.


poornas:
@donatello and @krishnasrinivas , pinging for reviews.

Minio/ApiEndpoints/ObjectOperations.cs
```diff
@@ -248,11 +248,17 @@ public partial class MinioClient : IObjectOperations
...
             int partNumber;
+            int numPartsUploaded = 0;
             bool skipUpload = false;
             for (partNumber = 1; partNumber <= partCount; partNumber++)
             {
                 byte[] dataToCopy = ReadFull(data, (int)partSize);
-
+                if (dataToCopy == null)
+                {
+                    skipUpload = true;
```
donatello: Looks like skipUpload is not read outside the for loop, so this line can be removed.

poornas: good catch, removed that line.

poornas:
@krishnasrinivas , this is waiting on your review.

PR #178, opened 2017-10-10
Added a new method SetUserMetadata to the PostPolicy

wowi42:
Hi there,

here is a PR for a new feature I need for a project I am working on with @hnb2.

# Feature

```SetUserSpecifiedMetadata(key string, value string)``` allows you to add user specified metadata following this guide which mention:

| key | explanation|
| ---  | ----------- |
| x-amz-meta-* | User-specified metadata.This condition supports exact matching and starts-with condition match type. |

### DIFF
--- a/Minio/DataModel/PostPolicy.cs
+++ b/Minio/DataModel/PostPolicy.cs
@@ -162,7 +162,7 @@ namespace Minio.DataModel
         /// </summary>
         /// <param name="status">Key and Value to insert in the metadata</param>
 
-        public void SetUserSpecifiedMetadata(string key, string value)
+        public void SetUserMetadata(string key, string value)
         {
             if (string.IsNullOrEmpty(key))
                 throw new ArgumentException("Key is Empty");
--- a/Minio/DataModel/PostPolicy.cs
+++ b/Minio/DataModel/PostPolicy.cs
@@ -158,6 +158,22 @@ namespace Minio.DataModel
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: Minio/DataModel/PostPolicy.cs [hunk 1]
```diff
@@ -158,6 +158,22 @@ public void SetSuccessStatusAction(string status)
...
+        public void SetUserSpecifiedMetadata(string key, string value)
+        {
+            if (string.IsNullOrEmpty(key))
+                throw new ArgumentException("Key is Empty");
+            if (string.IsNullOrEmpty(value))
+                throw new ArgumentException("Value is Empty");
+            string headerName = "x-amz-meta-" + key;
+            this.policies.Add(new Tuple<string, string, string>("eq", "$" + headerName, value));
+            this.formData.Add(headerName, value);
+        }
```
poornas: The SetKey() method already allows the user to set form data, and the custom metadata can be passed in with the x-amz-meta prefix. Is there a reason why an additional method would help

harshavardhana: It is coming from here https://github.com/minio/minio-go/pull/839

harshavardhana:
> SetUserSpecifiedMetadata(key string, value string) allows you to add user specified metadata following this guide which mention:

SetUserMetadata(key string, value string) is a better API IMO.  Borrowing suggestion by @vadmeste on minio-go PR.

wowi42:
Done

PR #184, opened 2017-11-01
fix: Add missing AWS region endpoints

### LINKED ISSUE #183: Regions (again) .netcore
Hi, seems that all regions other than us-east-1 are not working for `PutObjectAsync,` and there is no way to set regions in any constructor or parameter.  Am I missing something?  When I try `PutObjectAsync`, I get:

`File Upload Error: Minio API responded with message=The authorization header is malformed; the region 'us-east-1' is wrong; expecting 'us-east-2'`

poornas:
Fixes #183 

A few of the region specific endpoints were missing on .net sdk. Updating as per http://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region

### DIFF
--- a/Minio/Regions.cs
+++ b/Minio/Regions.cs
@@ -13,7 +13,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
+using System;
+using System.Text.RegularExpressions;
 namespace Minio
 {
     public class Regions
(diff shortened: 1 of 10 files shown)


Minio/Regions.cs
```diff
@@ -29,14 +29,22 @@ private Regions()
         /// <returns>Region corresponding to the endpoint. Default is 'us-east-1'</returns>
         internal static string GetRegion(string endpoint)
         {
-            if (endpoint.EndsWith("s3 - ap - northeast - 1.amazonaws.com"))
+            if (endpoint.EndsWith("s3-ap-northeast-1.amazonaws.com"))
```
krisis: Can't we get the region from the endpoint string without having to use EndsWith method to pattern match individual regions?

poornas: yes, updated PR.

poornas:
@krisis  and @vadmeste,this is waiting on your feedback.

poornas:
@harshavardhana and @krishnasrinivas , this is waiting for review

PR #192, opened 2017-11-25
Fix json logging for Mint

nitisht:
- Remove JSON indented formatting, as other SDKs do not use formatting
- Remove console.WriteLine, so JSON logs can be parsed

This is needed for https://github.com/minio/mint/pull/64

### DIFF
--- a/Minio.Functional.Tests/FunctionalTest.cs
+++ b/Minio.Functional.Tests/FunctionalTest.cs
@@ -226,7 +226,7 @@ namespace Minio.Functional.Tests
         }
         private static void runQuickTests(MinioClient minioClient)
         {
-              // Check if bucket exists
+            // Check if bucket exists
             BucketExists_Test(minioClient).Wait();
 
             // Create a new bucket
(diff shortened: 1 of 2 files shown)


Minio.Functional.Tests/FunctionalTest.cs
```diff
@@ -603,12 +601,12 @@ private static void runQuickTests(MinioClient minioClient)
                         await TearDown(minio, bucketName);
                 }
                 new MintLogger("PutObject_Test8","Tests whether PutObject with unknown stream-size passes",TestStatus.PASS,(DateTime.Now - startTime)).Log();
+            }
+            catch (Exception ex)
+            {
+                new MintLogger("PutObject_Test8","Tests whether PutObject with unknown stream-size passes",TestStatus.FAIL,(DateTime.Now - startTime),"",ex.Message, ex.ToString()).Log();
+            }
         }
-        catch (Exception ex)
```
poornas: this exception for the outer try block is intended to catch Setup_Test failure as well as the inner using {} block. So it shouldn't be removed.

nitisht: In this case I have just changed the indentation here @poornas block is not removed..

nitisht:
changing reviewers as @donatello it not available currently

Issue #202, opened 2017-12-20
Error response 'Difference between request time and server time'


babedev:
I integrated minio-dotnet with my dotnet core project. Everything work fine on Kestrel but it got error on IIS.

It responded 'Minio API responded with message=The difference between the request time and the server's time is too large.'

poornas:
could you run date command on both the machines - IIS server running minio-dotnet sdk and the other ubuntu server with minio running on it? You might want to check the globalization .NET settings on IIS server

babedev:
on Windows server: `Thursday, December 21, 2017 10:53:37 AM`
on Ubuntu server: `Thu Dec 21 10:53:37 +07 2017`

Both server use same timezone

poornas:
could you also log the system date inside your client app and see if there is any discrepancy with what the iis server shows? The request is signed with system date in utc format and no other date manipulation is done.

poornas:
@babedev, were you able to check if system date used in dotnet app matches iis server date and timezone?

babedev:
I got different time on IIS log. It showed 5PM but it should be 12AM. I think it is IIS time setting problem. I will keep you update when I got more information.

poornas:
thanks for the update @babedev. I am not familiar with IIS server, but looks like iisreset or changing [.net globalization settings](https://technet.microsoft.com/en-us/library/cc725971(v=ws.10).aspx) might help

poornas:
@babedev, is your issue resolved? Please let us know if we can assist further.

babedev:
I try to fix date time on IIS but I cannot find .NET globalisation on IIS manager. May be because of .NET core. I am trying other ways to change it in StartUp.cs.

babedev:
I found root cause of this problem. I think it is locale problem because IIS server send date time with Buddhist calendar system, but on Minio server use Gregorian calendar.

I will try fix calendar system on IIS side.

kannappanr:
Closing the issue as fixing calendar on IIS side will solve as the reporter has mentioned.

PooriaShariatzadeh:
I do have this problem too. changing .NET Globalization does not work and also changing the internet time and local time of the server didn't work. please help me to resolve this problem its take my time for too long.

PooriaShariatzadeh:
finally, I added the code below before starting a new client and my problem is go away 
`           var current = new CultureInfo("fa-IR");
            current.DateTimeFormat = new DateTimeFormatInfo();
[...]

PR #205, opened 2018-01-05
Convert date to Gregorian system

babedev:
According to #202 , I have problem with calendar system. I checked Minio server that it always check with Gregorian system. So I think it would be better if we convert date to Gregorian system before put them into headers.


poornas:
@babedev ,DateTime internally uses gregorian calendar. The sdk needs to be agnostic to any specific locale/culture info. imo, this PR is not necessary.

kannappanr:
Closing this PR based on comments from @poornas .

PR #206, opened 2018-01-09
fix: remove assertions from functional test

poornas:
This is wr.t issue https://github.com/minio/mint/issues/249. 

The mint log doesn't capture actual exception because of the Assert.Fail() statements - remove these so that the exceptions can be cascaded up to the logger. 

Note - this is not a fix for above issue, only to help unmask the error in mint logs for easier debugging.

### DIFF
--- a/Minio.Functional.Tests/FunctionalTest.cs
+++ b/Minio.Functional.Tests/FunctionalTest.cs
@@ -405,7 +405,6 @@ namespace Minio.Functional.Tests
             catch (MinioException ex)
             {
                 new MintLogger("MakeBucket_Test4",makeBucketSignature,"Tests whether MakeBucket with region and bucketname with . passes",TestStatus.FAIL,(DateTime.Now - startTime),"",ex.Message, ex.ToString(),args).Log();
-                //Assert.Fail();
             }
         }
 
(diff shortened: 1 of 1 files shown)


Minio.Functional.Tests/FunctionalTest.cs
```diff
@@ -977,9 +974,9 @@ private static void runQuickTests(MinioClient minioClient)
                     await minio.CopyObjectAsync(bucketName, objectName, destBucketName, destObjectName, conditions);
 
                 }
-                catch (MinioException)
+                catch (MinioException ex)
                 {
-                    Assert.Fail();
+                    throw ex;
```
ebozduman: This is more like a generic approach question/comment:
If we `catch` and simply `throw` the caught exception without adding our customized, more explanatory error message, then what is the point of catching these exceptions?
Wouldn't the existing code and approach be the same with not doing anything and letting the system built-in exception mechanism do its job and throw these exceptions on its own without our `catch` logic get in the way?

poornas: our objective here is to capture the exception in mint logger, which will be served by cascading error up the stack to calling function - the assert statement just causes mint to crash with a generic aggregate exception.

Minio.Functional.Tests/FunctionalTest.cs
```diff
@@ -752,7 +751,6 @@ private static void runQuickTests(MinioClient minioClient)
             }
             catch (Exception ex)
             {
-                Assert.Fail();
                 throw ex;
             }
             return;
```
donatello: My question is similar to @ebozduman's - doesn't throwing and catching an exception (especially of the base `Exception` type) have absolutely no effect? I am not too familiar with C#, but to illustrate my point in Python:

```py
def g():
    try:
        return f()
    except Exception as e:
        raise e

# Now calling f() and g() have exactly the same behaviour in terms 
# of values returned and exceptions thrown.
```

If this this the case, then all blocks like:

```
catch (Exception ex)
{
    throw ex
}
````
may be removed and it will have no effect on the program behaviour.

poornas: except in cases where we are intercepting OperationCanceledException, removed the unnecessary rethrowing of error.

Issue #214, opened 2018-03-09
"PresignedGetObjectAsync" method does not respect indicated Header parameters.


reinaldocoelho:
I'm trying to get a URL with temporary access with the PresignedGetObjectAsync () method to the stored object, and I intend to send the content-disposition as a parameter so that the file is downloaded with a different name from the one stored in Storage.

Upon request, a valid URL is received, which correctly downloads the file, but with the name that is in the Storage, and not with the name indicated in the Header.

The code that I am using to make the request to Mínio is:

```csharp
var _minioClient = new MinioClient(EndPoint, AccessKey, SecretKey);
var bucketName = "test";
var objectName = "folder1/testObject/abcdefgh.txt";
var expiresInt = 900;
var reqParams = new Dictionary<string, string>
{
	//{ "content-disposition", $"attachment;filename=MyDocument.txt;" },
	{ "response-content-disposition", $"inline;filename=MyDocument.txt;" },
};
return await _minioClient.PresignedGetObjectAsync(bucketName, objectName, expiresInt, reqParams).ConfigureAwait(false);
```

This returns the URL valid, but download file with name "abcdefgh.txt" instead of "MyDocument.txt".

reinaldocoelho:
Hi @poornas !
I created a PullRequest to include Request.Parameters in final URL.
Can you validate and give me a feedback?

The PR is #215.

PR #215, opened 2018-03-13
Update to use request.Parameters to form PresignGetUrl.

reinaldocoelho:
The current version of the SDK is not using the parameters included in the Request dictionary for the PresignedGetObjectAsync () function. As raised in issue #214 .

I set the V4Authenticator class to consider the parameters when mounting the URL.

I would like to ask for support for code validation and if necessary indicate where and how I should create unit or integration tests.

I hope it helps.

### DIFF
--- a/Minio/V4Authenticator.cs
+++ b/Minio/V4Authenticator.cs
@@ -277,11 +277,7 @@ namespace Minio
                 + expires
                 + "&";
             requestQuery += "X-Amz-SignedHeaders=host";
-
-            foreach(Parameter requestParam in request.Parameters)
-            {
-                requestQuery += $"&{requestParam.Name}={Uri.EscapeDataString((string)requestParam.Value)}";
-            }
+            requestQuery += string.Join("&", request.Parameters.Select(rp => $"{rp.Name}={Uri.EscapeDataString((string)rp.Value)}"));
 
             string canonicalRequest = GetPresignCanonicalRequest(client, request, requestQuery);
             byte[] canonicalRequestBytes = System.Text.Encoding.UTF8.GetBytes(canonicalRequest);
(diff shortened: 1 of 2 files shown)


luizcarlosfaria:
@reinaldocoelho maybe adding unit tests to this this scenario improve the coverage. It guarantee that all break changes, on this functional requirement, cause broken Continuous Integration.

Minio/V4Authenticator.cs
```diff
@@ -278,6 +278,11 @@ public string PresignURL(IRestClient client, IRestRequest request, int expires)
                 + "&";
             requestQuery += "X-Amz-SignedHeaders=host";
 
+            foreach(Parameter requestParam in request.Parameters)
```
AlbertoMonteiro: Refatoring sugestion:
```
requestQuery += string.Join("&", request.Parameters.Select(rp => $"{rp.Name}={Uri.EscapeDataString((string)rp.Value)}"));
```

poornas:
@reinaldocoelho , you could use utils.UrlEncode() instead of Uri.EscapeDataString method. The presigned url you generate is missing & before the user headers

poornas:
@reinaldocoelho, The request params cannot be concatenated to the url the way you have done - reason is the aws authorization header is computed from the canonical request which has a specific [format](https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html). For example, newline is required between query string and canonical headers - this would be missed by concatenating request headers prematurely to the url. You could review my [PR](https://github.com/minio/minio-dotnet/pull/216) and validate if you are able to download the object with correct attachment name. Thanks for all the effort you put into this - appreciate it!

reinaldocoelho:
Fixed on #216.

PR #216, opened 2018-03-13
Fix presign operations to honor request params

### LINKED ISSUE #214: "PresignedGetObjectAsync" method does not respect indicated Header parameters.

poornas:
Fixes #214

### DIFF
--- a/Minio/V4Authenticator.cs
+++ b/Minio/V4Authenticator.cs
@@ -278,7 +278,9 @@ namespace Minio
                 + "&";
             requestQuery += "X-Amz-SignedHeaders=host";
 
-            string canonicalRequest = GetPresignCanonicalRequest(client, request, requestQuery);
+            SortedDictionary<string,string> headersToSign = GetHeadersToSign(request);
+            string canonicalRequest = GetPresignCanonicalRequest(client, request, requestQuery, headersToSign);
+            string headers = string.Join("&", headersToSign.Select(p => p.Key + "=" + utils.UrlEncode(p.Value)));
             byte[] canonicalRequestBytes = System.Text.Encoding.UTF8.GetBytes(canonicalRequest);
             string canonicalRequestHash = BytesToHex(ComputeSha256(canonicalRequestBytes));
             string stringToSign = GetStringToSign(this.Region, signingDate, canonicalRequestHash);
@@ -288,7 +290,7 @@ namespace Minio
             string signature = BytesToHex(signatureBytes);
 
             // Return presigned url.
-            return client.BaseUrl + path + "?" + requestQuery + "&X-Amz-Signature=" + signature;
+            return client.BaseUrl + path + "?" + requestQuery + "&" + headers + "&X-Amz-Signature=" + signature;
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


Minio/V4Authenticator.cs
```diff
@@ -310,7 +312,12 @@ private string GetPresignCanonicalRequest(IRestClient client, IRestRequest reque
                 path = "/" + path;
             }
             canonicalStringList.AddLast(path);
-            canonicalStringList.AddLast(requestQuery);
+            String query = requestQuery;
+            foreach (string header in headersToSign.Keys)
```
AlbertoMonteiro:
```csharp
requestQuery += string.Join("", headersToSign.Select(h => $"&{h.Key}={utils.UrlEncode(h.Value)}"));
```

Minio/V4Authenticator.cs
```diff
@@ -310,7 +312,12 @@ private string GetPresignCanonicalRequest(IRestClient client, IRestRequest reque
...
+            {
+               query = query + "&" + header + "=" + utils.UrlEncode(headersToSign[header]);
+            }
```
AlbertoMonteiro: You are looping through dictionary, and then getting the value using the indexed acessor(that is a overhead), my suggestion does not get the overhead because it use the property value, from current dictionary entry.
```csharp
var query = headersToSign.Aggregate(requestQuery, (pv, cv) => $"{pv}&{h.Key}={utils.UrlEncode(h.Value)}");
```

poornas: Thanks @AlbertoMonteiro, that's more concise.

poornas:
@krisis, we have functional tests that test both presign url generation and perform a request with said URL - have updated functional tests to assert if response headers show that the request headers were honored.

reinaldocoelho:
Do you have a release date forecast with this fix?

poornas:
@reinaldocoelho, waiting for another PR to be merged before making  a release. Hopefully by end of day.

reinaldocoelho:
It's wonderful.
Thank's.

poornas:
@reinaldocoelho ,new version of .net sdk has been released with the fix.

reinaldocoelho:
Using on my Dev branch.
Thanks 👍

PR #220, opened 2018-06-02
Functionaltest: fix missing content length header

poornas:
Test is setting policyjson to an empty string, causing GetBucketPolicy test to fail with Content-Length header missing.


poornas:
@krisis, removed the truth assertion for this test - the policy json stored at the server can come back in a slightly different format than the policy json string that was sent. Hence omitting that assertion, since the call is more or less a Get/Set.

PR #221, opened 2018-06-05
fix url encoding for presign urls

poornas:
When accesskey contains special characters, it causes signature
mismatch error.

fixes https://github.com/minio/minio/issues/6011

### DIFF
--- a/Minio/V4Authenticator.cs
+++ b/Minio/V4Authenticator.cs
@@ -267,8 +267,8 @@ namespace Minio
 
             requestQuery = "X-Amz-Algorithm=AWS4-HMAC-SHA256&";
             requestQuery += "X-Amz-Credential="
-                + this.accessKey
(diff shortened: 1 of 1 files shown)


Minio/V4Authenticator.cs
```diff
@@ -267,7 +267,7 @@ public string PresignURL(IRestClient client, IRestRequest request, int expires)
 
             requestQuery = "X-Amz-Algorithm=AWS4-HMAC-SHA256&";
             requestQuery += "X-Amz-Credential="
-                + this.accessKey
+                + Uri.EscapeDataString(this.accessKey)
```
harshavardhana: It is better to do this perhaps across the entire string. like 

```
Uri.EscapeDataString(this.accessKey + "/" + GetScope(...))
```

poornas:
@harshavardhana , updated PR.

PR #223, opened 2018-06-08
fix: pass region down correctly to authenticator

poornas:
Fixes https://github.com/minio/minio/issues/5995


poornas:
@krisis and @harshavardhana , PTAL

PR #233, opened 2018-07-06
API: Change CopyObjectAsync to support metadata replacement

### LINKED ISSUE #229: Support setting metadata for CopyObject
Hello,

We are looking for a way to change the metadata of an object stored in minio without downloading and reuploading the object. I've learned that S3 does support changing metadata by using the CopyObject and using the same bucket and object-key and it also seems minio does support that. See here:
https://github.com/minio/minio/issues/3316

The .NET SDK does not support metadata for CopyObject.
Is it possible you're going to support that?

Thank you.

poornas:
on server side copy

fixes #229


Minio.Examples/Cases/CopyObjectMetadata.cs
```diff
@@ -0,0 +1,62 @@
+﻿/*
+ * Minio .NET Library for Amazon S3 Compatible Cloud Storage, (C) 2017 Minio, Inc.
```
krisis: copyright should include 2018

PR #236, opened 2018-08-07
Fix .NET Core host header generation if using default ports

WolfspiritM:
When using .NET Core and a reverse proxy with default ports, then the signed Headers create a header with "domain.com" instead of "domain.com:443" when using .NET Core. It seems the Server however always includes the Port. This results in a signature mismatch.

I can't see why this should be different for .NET Core and .NET Framework.

For .NET 4.6.2 it produces the following request that is signed:
```PUT\n/test\n\nhost:domain.com:443\nx-amz-content-sha256:....```
For Core it generates it like that:
```PUT\n/test\n\nhost:domain.com\nx-amz-content-sha256:....```

This PR removes the switch between Core and Framework and handles everything like the Framework version.

### DIFF
--- a/Minio/V4Authenticator.cs
+++ b/Minio/V4Authenticator.cs
@@ -414,30 +414,10 @@ namespace Minio
             {
                 string headerName = header.Name.ToLower();
                 string headerValue = header.Value.ToString();
-#if NET46
                 if (!ignoredHeaders.Contains(headerName))
                 {
                     sortedHeaders.Add(headerName, headerValue);
                 }
-#else
-                if (headerName.Equals("host"))
-                {
-                    var host = headerValue.Split(':')[0];
-                    var port = headerValue.Split(':')[1];
-                    if (port.Equals("80") || port.Equals("443"))
-                    {
-                        sortedHeaders.Add(headerName, host);
-                    }
-                    else
-                    {
-                        sortedHeaders.Add(headerName, headerValue);
-                    }
-                }
-                else if (!ignoredHeaders.Contains(headerName))
-                {
-                    sortedHeaders.Add(headerName, headerValue);
-                }
-#endif
             }
 
             return sortedHeaders;

### CODE REVIEW

File: Minio/V4Authenticator.cs [hunk 1]
```diff
@@ -414,30 +414,10 @@ private string GetPresignCanonicalRequest(IRestClient client, IRestRequest reque
...
-                    else
-                    {
-                        sortedHeaders.Add(headerName, headerValue);
-                    }
-                }
-                else if (!ignoredHeaders.Contains(headerName))
-                {
-                    sortedHeaders.Add(headerName, headerValue);
-                }
-#endif
```
poornas: @WolfspiritM, standard ports such as 80 and 443 are removed from the host header before generating authorization header - this is correctly handled in the .netcore specific code. 

W.r.t your presigned request for .netcore failing with signature mismatch, can you paste the actual error message from minio and the reverse proxy config?

WolfspiritM: Why is that only handled in .NET Core then and not in .NET Framework? Both generate a different signature and with .NET Framework the connection worked while with .NET Core I got the Mismatch error.

```
Minio API responded with message=The request signature we calculated does not match the signature you provided. Check your key and signing method.'
```

I'm using traefik as reverse proxy. I now started minio with --address :80 and the following .NET core code:
``` c#
 var minioClient = new MinioClient("localhost",
                "HX4OBAN38VW39BYB5KS5",
                "ydmDD0AqRPTfqV0ZbAmsgMJl6+G2gfFkb9wKcuyc");

            var t = minioClient.BucketExistsAsync("test").GetAwaiter().GetResult();
            Console.WriteLine(t);
            Console.ReadLine();
```
and that fails aswell for .NET Core but works for .NET Framework. (just switching the Framework to net461 fixes it) so it's not a reverse proxy problem!

The NetCore version works fine for .NET Core. Only the new 2.0 version breaks and while debugging for .net framework and .net core I noticed that .net framework adds the port and .net core does not. After changing it the way described in this PR it works for both ways.

WolfspiritM: @poornas ```<TargetFramework>netcoreapp2.1</TargetFramework>```

```
.NET Core SDK (reflecting any global.json):
 Version:   2.1.301
 Commit:    59524873d6
```

PR #237, opened 2018-08-09
Fix Uploading QR Codes

Ashkan-rmk:
When you generate QR Code and want to upload it to storage, it uploaded with zero length.

with this method, you can convert stream to byte array easily with any kind of stream generator.

### DIFF
--- a/Minio/ApiEndpoints/ObjectOperations.cs
+++ b/Minio/ApiEndpoints/ObjectOperations.cs
@@ -209,7 +209,7 @@ namespace Minio
             {
                 contentType = "application/octet-stream";
             }
-            if (! metaData.ContainsKey("Content-Type"))
+            if (!metaData.ContainsKey("Content-Type"))
             {
                 metaData["Content-Type"] = contentType;
             }
@@ -261,7 +261,7 @@ namespace Minio
                 {
                     break;
                 }
-                
+
                 if (partNumber == partCount)
                 {
                     expectedReadSize = lastPartSize;
@@ -431,7 +431,7 @@ namespace Minio
         /// <param name="metaData"></param>
         /// <param name="cancellationToken">Optional cancellation token to cancel the operation</param>
         /// <returns></returns>
-        private async Task<string> NewMultipartUploadAsync(string bucketName, string objectName, Dictionary<string,string> metaData, CancellationToken cancellationToken)
+        private async Task<string> NewMultipartUploadAsync(string bucketName, string objectName, Dictionary<string, string> metaData, CancellationToken cancellationToken)
         {
             var resource = "?uploads";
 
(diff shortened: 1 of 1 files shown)


Minio/ApiEndpoints/ObjectOperations.cs
```diff
@@ -842,33 +842,13 @@ public async Task<ObjectStat> StatObjectAsync(string bucketName, string objectNa
         /// <returns>bytes read in a byte array</returns>
```
harshavardhana: The logic of currentPartSize is totally ignored here @Ashkan-rmk you need to see the flow why ReadFull() is implemented. It is to upload large files with an optimal part size. Your code simply loads everything into memory it will crash the program. 

What you need to do is simply provide the stream and with putObjectAsync taking in -1 for length as the input parameter.

Ashkan-rmk: I use QR code generator nuget package. It generate a memory stream object and when I pass it to putObjectAsync it uploaded with 0 byte!

you mean that I have to pass "-1" as a parameter to putObjectAsync?

poornas: that's correct. pass size as -1

Ashkan-rmk: I pass size as -1 but it returns 

The XML you provided was not well-formed or did not validate against our published schema.

what should I do?

Ashkan-rmk: public static void Main(string[] args)
        {
            string accessKey = "xxxx";
            string secretKey = "xxxxxxxxx";
            string ServiceURL = "miniostage.sloppy.zone";

            client = new MinioClient(ServiceURL, accessKey, secretKey).WithSSL();

            try
            {
                UploadImageAsync().Wait();
            }
            catch (Exception ex)
            {
                Console.WriteLine($"There was an exception: {ex.ToString()}");
            }

        }

        public static async Task UploadImageAsync()
        {
            string bucketName = "files";

            var qr_code = PayPing.Common.Tools.QrCodeGeneratorHelper.GenerateUrl("https://payping.ir/d/352", true, false);
            var extension = qr_code.imageFormat.GetImageFileExtension().Replace(".", "");

            string keyName = $"QRTest23.{extension}";

            await client.PutObjectAsync(
                      bucketName,
                      keyName,
                      qr_code.stream,
                      -1,
                      "application/octet-stream");

            Console.WriteLine(keyName);
        }

Ashkan-rmk: the QR Code successsfuly generate stream.
as you said I pass -1 as parameter to PutObjectAsync, but it throw this exception.

> The XML you provided was not well-formed or did not validate against our published schema.

krisis: @Ashkan-rmk I replaced the QR code generation with a simple `Byte` array based `MemoryStream` with rest of the program as is. I don't see the above exception and the object is uploaded successfully. Could you come on https://slack.minio.io/? It would be easier to resolve the issue over slack.

poornas: @Ashkan-rmk, would suggest copying your stream to a memory stream and pass size as -1 for the PutObjectAsync call.

Ashkan-rmk: I came to Slack channel... thanks.

PR #242, opened 2018-09-10
Remove unnecessary Newtonsoft

SanSYS:
I have a little problem - big solution, with references to Newtonsoft Json 9 version
And I wanto to use your library
But reference to Newtonsoft version of 11 breaks my project
After review your code I understood, that Newtonsoft uses only for tracing to console
Not for production ;)
So.. I changed it to StringBuilder

image

And I think build/deploy will be faster ;)

### DIFF
--- a/Minio.Functional.Tests/MintLogger.cs
+++ b/Minio.Functional.Tests/MintLogger.cs
@@ -16,7 +16,7 @@
 
 using System;
 using System.Collections.Generic;
-using Newtonsoft.Json;
+using System.Text;
 
 namespace Minio.Functional.Tests
 {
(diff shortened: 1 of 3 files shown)


nitisht:
@SanSYS Thanks for the PR, could you please check the TravisCI failure?

SanSYS:
i'm not sure how to resolve the problem... 
I'll try rerun all tests in my machine later

image

nitisht:
> i'm not sure how to resolve the problem...

no problem, let me try re-running the tests on Travis

poornas:
@SanSYS, the functional tests output needs to be in json format, as our automated tests require it. As an alternative JavascriptSerializer can be considered - but this is compatible with .NET framework  but not .NetCore. This is why we have Newtonsoft JSON as a project dependency to be able to have JSON output for both .NET framework and .NET Core

SanSYS:
@poornas thank you for your answer, I'll suggest solution later

PR #243, opened 2018-09-10
Some issues

SanSYS:
And I think you have some problems with async/await methods
Please - try to use sonar qube or linter for your IDE https://www.sonarlint.org/


SanSYS:
so? maybe it's time to merge? )

PR #248, opened 2018-09-20
Remove unnecessary Newtonsoft from main package

SanSYS:
I have a little problem - big solution, with references to Newtonsoft Json 9 version
And I want to use your library
But reference to Newtonsoft version of 11 breaks my project
After review your code I understood, that Newtonsoft used only for tracing to console
Not for production ;)

So.. I created interface `IRequestLogger` with default implemetation via `StringBuilder`
And if anybody wants to use newtonsoft, csv or yet another logger - he could create himself implemetation )

### DIFF
--- a/Minio.Functional.Tests/MintLogger.cs
+++ b/Minio.Functional.Tests/MintLogger.cs
@@ -14,9 +14,9 @@
 * limitations under the License.
 */
 
+using Newtonsoft.Json;
 using System;
 using System.Collections.Generic;
-using System.Text;
 
 namespace Minio.Functional.Tests
 {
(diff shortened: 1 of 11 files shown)


SanSYS:
@poornas is it ok now?
Or should I do anymore?

Minio.Functional.Tests/MintLogger.cs
```diff
@@ -77,10 +77,37 @@ public MintLogger(string testName,string function, string description,TestStatus
...
+                    sb.Append(kv.Key);
+                    sb.Append("]=");
+
+                    sb.AppendLine(kv.Value);
+                }
+            }
+
 
-          Console.Out.WriteLine(JsonConvert.SerializeObject(this,Formatting.None,
-            new JsonSerializerSettings { NullValueHandling = NullValueHandling.Ignore }));
```
poornas: @SanSYS, since Minio.FunctionalTests.csproj is retaining the NewtonSoft.JSON dependency, you can leave the Log() function code as it was, i.e. we just want lines 82 and 83 that you removed. The functional tests output needs to be in json format.
```
{"name":"minio-dotnet: BucketExists_Test","function":"Task<bool> BucketExistsAsync(string bucketName, CancellationToken cancellationToken = default(CancellationToken))","args":{"bucketName":"miniodotnetkxu3g"},"duration":2477,"status":"PASS"}
```

SanSYS: Ok, thanks!

SanSYS:
moved into https://github.com/minio/minio-dotnet/pull/249

PR #266, opened 2018-12-28
Use region supplied in MinioClient constructor while sending requests

krisis:
Thanks to @AdmirR for pointing out the appropriate V4Authenticator constructor in CreateRequest.


krisis:
@AdmirR, could you review this PR? I tested this PR locally by setting region as "us-west-1" (non default) for a local Minio object server and instantiating MinioClient object with "us-west-1" as region parameter.

AdmirR:
@krisis Tested and works now, thanks!

PR #268, opened 2019-01-04
Fix case-sensitive comparison for `ETag`

MarcosBidtellect:
We were having issues where sometimes the `ETag` header parameter would come as `Etag`, which would cause `ObjectStat.ETag` to be an empty string.

Making the string comparison case-insensitive fixed the issue.


harshavardhana:
All HTTP headers interpreted should be case insensitive @poornas - we may need to check if we have more places such as these.

poornas:
will do @harshavardhana

kannappanr:
@harshavardhana Will merge this PR now. @poornas will be sending in another PR to fix the remaining ones.

PR #276, opened 2019-01-18
Allow HTTP Proxy to be configured

### LINKED ISSUE #273: HTTP Proxy support
Hi,

Last I checked, I wasn't able to locate any way to configure an HTTP proxy. Would it be possible to support this?

We have a use case where our software is run in arbitrary locations that may or may not be reachable by end users. If it's not reachable, there will usually be some form of proxy set up that we should use. 

Mike.

poornas:
fixes #273

### DIFF
--- a/Docs/API.md
+++ b/Docs/API.md
@@ -62,6 +62,11 @@ __Secure Access__
 |---|
 |`Chain .WithSSL() to Minio Client object to use https instead of http. `   |
 
+__Proxy__
+
+|  |
+|---|
+|`Chain .WithProxy(proxyObject) to Minio Client object to use proxy `   |
 
 __Example__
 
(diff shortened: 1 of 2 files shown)


krisis:
@poornas,
1.  Why are we not adding a new (overloaded) constructor which takes the proxy as a parameter?
     If it is exposed as a member of `MinioClient` object, isn't it susceptible to data race like proxy being modified while a PutObject of a large object is ongoing? i.e UploadPart API calls may be sent over the proxy while the CompleteMultipartUpload may be sent over the updated proxy. This causes unnecessary confusion.
2.  We should update the README with a section describing how and when to use the proxy setting in `MinioClient`.

poornas:
@krisis, Proxy is set only once during initialization of the underlying RestSharp client, and hence the data race scenario you stated above will not arise. This doesn't have to be added to README imo, because it is not per se a minio-dotnet sdk feature.

krisis:
@poornas, I understand why there won't be any data race like I thought earlier. Thanks for clarifying that. Won't it help our users if we added an example that demonstrates how and when to set proxy client?

poornas:
@krisis, added an example and changed it to a chained constructor.

sinhaashish:
LGTM

PR #287, opened 2019-02-19
Reduce filesizes used in functional tests

poornas:
to speed up Mint.Also fixing metadata dictionary to be
case insensitive.

### DIFF
--- a/Minio.Functional.Tests/FunctionalTest.cs
+++ b/Minio.Functional.Tests/FunctionalTest.cs
@@ -32,8 +32,12 @@ namespace Minio.Functional.Tests
     public class FunctionalTest
     {
         private static Random rnd = new Random();
+        private static int KB = 1024;
         private static int MB = 1024 * 1024;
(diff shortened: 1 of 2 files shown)


Minio.Functional.Tests/FunctionalTest.cs
```diff
@@ -2126,13 +2132,9 @@ private async static Task RemoveObjects_Test2(MinioClient minio)
                 {
                     tasks[i] = PutObject_Task(minio, bucketName, objectName + i.ToString(), null, null, 0, null, rsg.GenerateStreamFromSeed(5));
                     objectsList.Add(objectName + i.ToString());
-                    // Add sleep to avoid flooding server with concurrent requests
```
kannappanr: Is this not required anymore?

poornas: no, because only 50 objects are being uploaded

PR #292, opened 2019-03-13
Fix DateTime parse exception

### LINKED ISSUE #291: DateTime.Parse Thrown Exception
Hi,  i'm use minio on a project as a object store 

When I try put some objecto on store the parsing of the response throws an exception on the `Last-Modified` HTTP hader:

"String was not recognized as a valid DateTime."

The reason seems to be the DateTime.Parse here:

https://github.com/minio/minio-dotnet/blob/c7857cc2904fcd71bfae60fe8a8108e8fc414bde/Minio/ApiEndpoints/ObjectOperations.cs#L805

My current culture is `es-AR`, so i solve the exception forcing the culture info to invariant in date parsing

`lastModified = DateTime.Parse(parameter.Value.ToString(), CultureInfo.InvariantCulture);`

thanks.

poornas:
Fixes #291

### DIFF
--- a/Minio/ApiEndpoints/ObjectOperations.cs
+++ b/Minio/ApiEndpoints/ObjectOperations.cs
@@ -17,6 +17,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using RestSharp;
 
 using System.Threading.Tasks;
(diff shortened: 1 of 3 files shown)


poornas:
> @poornas, the changes look good to me. Would it make sense to have a subtype to the standard `DateTime`, say like `MinioDateTime`, whose `Parse` method will internally use `CultureInfo.InvariantCulture` like in this PR. I suggest this,
> 
>     1. to avoid repetition in all the places in code where we use `DateTime`
> 
>     2. to make it a default across our SDK.
@krisis, we parse DateTime only in these 3 instances.Doesn't seem to warrant a separate data model since the language provides a convenient overloaded method for parsing DateTime invariant of the culture. What do you think?

PR #302, opened 2019-05-04
 Cleanup, fix async, optimize format string

tinohager:
Code cleanup, and some small fixes

PR #303, opened 2019-05-06
Code cleanup

sungam3r:
Code cleanup


sungam3r:
Just saw that there is another pull request #302 for code cleanup. Apparently, the project required it.

Issue #308, opened 2019-06-03
Sign Minio dll


adityayenduri:
Hello Team,

Can you please sign the minio dll? As our project has strongkey and its not allowing to add minio dll.

Thanks in advance,
Aditya Y

harshavardhana:
We will take a look @adityayenduri - Thanks for opening an issue.

PR #324, opened 2019-08-06
Add strong signing to assembly

poornas:
Fixes: #308.

Version being bumped to 3.1.0


poornas:
> Should Minio.Snk file be removed?

@kannappanr , this is the [official recommendation](https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/strong-naming#create-strong-named-net-libraries) on how to handle the .snk files

PR #337, opened 2019-09-23
Unescape URI encoded returns in listObjectsResult

harshavardhana:
8bd4c65 added encoding-type
this broke recursive listing for special characters due to
wrong marker calculation, this PR fixes this.

### DIFF
--- a/Minio.Functional.Tests/FunctionalTest.cs
+++ b/Minio.Functional.Tests/FunctionalTest.cs
@@ -85,6 +85,18 @@ namespace Minio.Functional.Tests
             return GetFilePath(dataFileName);
         }
 
+        public static string GetRandomObjectName(int length = 5)
+        {
+            string characters = "abcd+&%$#@*&{}[]()";
+            StringBuilder result = new StringBuilder(length);
+
+            for (int i = 0; i < length; i++)
+            {
+                result.Append(characters[rnd.Next(characters.Length)]);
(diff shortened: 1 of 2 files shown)


krisis:
@harshavardhana do you have a test case to recreate the issue this PR is fixing? Does this problem happen in minio-go too after https://github.com/minio/minio-go/pull/1161?

harshavardhana:
> @harshavardhana do you have a test case to recreate the issue this PR is fixing? Does this problem happen in minio-go too after [minio/minio-go#1161](https://github.com/minio/minio-go/pull/1161)?

Yes and here it is 

### To reproduce

```
~ for i in $(seq 1 1001); do mc cp 1++1 myminio-local1/testbucket/prefix/1++1-${i}; done
```

```
~ mc ls myminio-local1/testbucket/prefix/
```

This will hang forever listing first 1000 entries repeatedly.

Minio.Tests/TestHelper.cs
```diff
@@ -23,6 +23,18 @@ class TestHelper
     {
         private static Random rnd = new Random();
 
+        public static string GetRandomObjectName(int length = 5)
```
poornas: this function should have been added in  `Minio.Functional.Tests/FunctionalTest.cs` - functional tests and unit tests are in different namespaces here.

Issue #342, opened 2019-10-08
Both spaces and pluses are serialized to pluses in the .NET SDK


sebastianbk:
Hello,

I am making use of the Minio SDK for .NET together with the [Minio Gateway for GCS](https://docs.min.io/docs/minio-gateway-for-gcs.html) and, in the beginning, everything was working flawlessly. I really appreciate how easy it is to get started with and how well it generally works.

However, I have come across an issue that's causing quite a lot of trouble for me. When using the `ListObjectsAsync` method in the .NET SDK, I have noticed that both spaces and pluses are being serialized to pluses, making it impossible to know what the real file/folder name is.

As I already have thousands of buckets with all kinds of filenames in them, I really can't seem to figure out whether a plus is really a space or a plus. I checked in the Minio Browser and, there, there is no issue. It distinguishes nicely between my files with spaces and pluses in them, respectively.

Is this a bug in the SDK or a result of communicating with the Minio Gateway over HTTP where spaces can sometimes be serialized as pluses?

I would very much like to offer my help in solving this issue, as I would love to make use of Minio but can't right now.

harshavardhana:
Yes we need to fix this.. Thanks for reporting @sebastianbk

vadmeste:
> Yes we need to fix this.. Thanks for reporting @sebastianbk

The issue should be fixed after releasing minio-go and updating it in minio/minio. I am happy to test gateway S3 & Azure to be sure.

harshavardhana:
Here it's gcs gateway - I think the issue is in dotnet SDK

vadmeste:
> Here it's gcs gateway - I think the issue is in dotnet SDK

My bad, yes.. GCS gateway doesn't use minio-go

sebastianbk:
Wauw, thanks for the fast response on this issue. Your commitment is really inspiring!

Would you be able to offer a timeline for when the fix could be released?

harshavardhana:
No timeline @sebastianbk but soon.

sebastianbk:
Let me know when you'll release this. I am desperate to make use of it! 😃

PR #343, opened 2019-10-08
Fix unescaping encoded characters in ListObjectsAsync API

poornas:
Fixes: #342.

[Uri.UnescapeDataString](https://docs.microsoft.com/en-us/dotnet/api/system.uri.unescapedatastring?view=netframework-4.8) does not convert plus characters into spaces while decoding, which poses issues as evidenced here. Instead use HttpUtility.UrlDecode()


heedfull:
I'm not sure what you mean. I applied the same fix as #343 and the new functional test passes only once the change was made.

There are no new failing tests after my change.

PR #362, opened 2019-12-26
Fix canned ACLs being treated as metadata

### LINKED ISSUE #361: Canned ACLs erroneously get prefixed with x-amz-meta
Hello,

I realize that the Minio client is designed for Minio, but I'm opening this issue because I believe this affects the S3 compatibility of the library. I have been using the Minio Client with DigitalOcean Spaces for over half a year now, and it's been working fantastic, until #356 which broke public accesss for `PutObjectAsync` on DO.

DigitalOcean Spaces has some special treatment for what they (and the [AWS docs](https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html)) call "canned" ACLs, one of which is the `x-amz-acl` metadata property, which indicates to DO that the object is publicly accessible. This used to work fine with the Minio client if I set the metadata to: `new Dictionary<string, string> {{"x-amz-acl", "public-read"}};` 

However, as of 3.1.8, this translates to `x-amz-meta-x-amz-acl | public-read`, which DO doesn't pick up on, and subsequently the object is kept private. 

I believe this to be a bug in the Minio client, because the S3 documentation mentions sending the `x-amz-acl` header, without the `x-amz-meta` prefix.

The following code reproduces this problem when used against DigitalOcean Spaces with Minio version 3.1.8:

```
class Program
{
    private static MinioClient _client;

    static async Task Main(string[] args)
    {
        _client = new MinioClient("DO_EndPoint", "DO_Key", "DO_Secret").WithSSL();

        await PutAsync("bucket", "someobject", "text", Enumerable.Range(0, 100).Select(i => (byte) i).ToArray(),
            true);

        Console.WriteLine("Hello World!");
    }

    public static async Task PutAsync(string bucketName, string objectName, string contentType, byte[] content,
        bool isPublic)
    {
        await using var ms = new MemoryStream(content);

        if (!await _client.BucketExistsAsync(bucketName))
            await _client.MakeBucketAsync(bucketName);

        Dictionary<string, string> metadata = null;
        if (isPublic)
            metadata = new Dictionary<string, string> { { "x-amz-acl", "public-read" } };

        await _client.PutObjectAsync(bucketName, objectName, ms, ms.Length, contentType, metadata);
    }
}
```

This can be worked around by staying on version 3.1.7.

aevitas:
Fixes #361 in the simplest manner I could find.

### DIFF
--- a/Minio.Functional.Tests/FunctionalTest.cs
+++ b/Minio.Functional.Tests/FunctionalTest.cs
@@ -446,8 +446,7 @@ namespace Minio.Functional.Tests
             string contentType = "custom/contenttype";
             var metaData = new Dictionary<string, string>
             {
-                { "customheader", "minio   dotnet" },
-                { "x-amz-acl", "public-read" }
+                { "customheader", "minio   dotnet" }
             };
             var args = new Dictionary<string, string>
             {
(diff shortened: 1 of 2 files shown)


aevitas:
What would be a good way to test this? I thought testing for the metadata in `PutObject_Test4` would be a good spot, but it appears the `StatObject` doesn't contain the metadata I passed when creating the object.

poornas:
> What would be a good way to test this? I thought testing for the metadata in `PutObject_Test4` would be a good spot, but it appears the `StatObject` doesn't contain the metadata I passed when creating the object.

ACL metadata is not returned in StatObject - the SDK needs a GetObjectACL to fetch ACL info. If this PR fixes your acl issue on DO spaces, you can remove the assert for ACL in PutObject_Test4 and we can approve this PR.

aevitas:
I can confirm this change fixes my issue on DO Spaces. I've reverted the tests commit and am waiting for CI to complete.

Thanks for getting back to me so quickly!

PR #367, opened 2020-01-09
Add SelectObjectContentAsync API call

### LINKED ISSUE #359: Select object content
Hi,
do you plan to add select object content to the .NET API as well? Is there already any timeline for that? Thank you for the info

poornas:
fixes #359

### DIFF
--- a/Minio/ApiEndpoints/ObjectOperations.cs
+++ b/Minio/ApiEndpoints/ObjectOperations.cs
@@ -1,6 +1,6 @@
 ﻿/*
  * MinIO .NET Library for Amazon S3 Compatible Cloud Storage,
- * (C) 2017, 2018, 2019 MinIO, Inc.
+ * (C) 2017, 2018, 2019, 2020 MinIO, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -180,6 +180,42 @@ namespace Minio
                 }
                 utils.MoveWithReplace(tempFileName, fileName);
             }, sse, cancellationToken).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Select an object's content. The object will be streamed to the callback given by the user.
+        /// </summary>
+        /// <param name="bucketName">Bucket to retrieve object from</param>
+        /// <param name="objectName">Name of object to retrieve</param>
+        /// <param name="opts">Select Object options</param>
+        /// <param name="cancellationToken">Optional cancellation token to cancel the operation</param>
+        public async Task<SelectResponseStream> SelectObjectContentAsync(string bucketName, string objectName, SelectObjectOptions opts,CancellationToken cancellationToken = default(CancellationToken))
+        {
+            utils.ValidateBucketName(bucketName);
+            utils.ValidateObjectName(objectName);
+            if (opts == null)
+            {
+                throw new ArgumentException("Options cannot be null", nameof(opts));
+            }
+            Dictionary<string,string> sseHeaders = null;
+            if (opts.SSE != null)
+            {
+                sseHeaders = new Dictionary<string,string>();
+                opts.SSE.Marshal(sseHeaders);
+            }
+            var selectReqBytes = System.Text.Encoding.UTF8.GetBytes(opts.MarshalXML());
+
+            string resourcePath = "?select=&select-type=2";
+            var request = await this.CreateRequest(Method.POST, bucketName,
+                                                    objectName: objectName,
+                                                    resourcePath: resourcePath,
+                                                    headerMap: sseHeaders)
+                                    .ConfigureAwait(false);
+
+            request.AddParameter("application/xml", selectReqBytes, ParameterType.RequestBody);
+
+            var response = await this.ExecuteTaskAsync(this.NoErrorHandlers, request, cancellationToken).ConfigureAwait(false);
+            return new SelectResponseStream(new MemoryStream(response.RawBytes));
         }
 
         /// <summary>
--- a/Minio/DataModel/Select/SelectResponseStream.cs
+++ b/Minio/DataModel/Select/SelectResponseStream.cs
@@ -1,0 +1,256 @@
+/*
+ * MinIO .NET Library for Amazon S3 Compatible Cloud Storage, (C) 2020 MinIO, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+using Minio.Exceptions;
+
+using System;
+using System.Collections.Generic;
+
+using System.IO;
+using System.Runtime.Serialization.Formatters.Binary;
+using System.Xml.Serialization;
+using RestSharp;
+using System;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Force.Crc32;
+using System.Linq;
+using System.Collections.Generic;
+
+namespace Minio.DataModel
+{
+    [Serializable]
+    public class SelectResponseStream
+    {
+        // SelectResponseStream is a struct for selectobjectcontent response.
+        
+        public Stream Payload  { get ; set; }  
+
+        [XmlElement("Stats", IsNullable = false)]
+        public StatsMessage Stats { get; set; }
+
+        [XmlElement("Progress", IsNullable = false)]
+        public ProgressMessage Progress { get; set; }
+
+        private byte[] prelude = new byte[8];
+        private byte[] preludeCRC = new byte[4];
+        private byte[] messageCRC = new byte[4];
+        private byte[] headerValueLen = new byte[2];
+
+        private MemoryStream payloadStream = null;
+        private bool _isProcessing;
+        
+        public SelectResponseStream()
+        {
+        }
+
+        protected int ReadFromStream(byte[] buffer)
+        {
+            int read = -1;
+            if  (!this._isProcessing)
+            {
+               return read;
+            }
+            read = this.payloadStream.Read(buffer, 0, buffer.Length);
+            if (!this.payloadStream.CanRead)
+            {
+                this._isProcessing = false;
+            }
+            return read;
+        }
+        public SelectResponseStream(Stream s)
+        {
+            if (s != null) {
+                var _ms = new MemoryStream();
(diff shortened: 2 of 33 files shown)


Minio/DataModel/Select/SelectResponseStream.cs
```diff
@@ -0,0 +1,249 @@
...
+                var n =  ReadFromStream(prelude);
+                numBytesRead += n;
+                n =  ReadFromStream(preludeCRC);
+                numBytesRead += n;
+                var inputArray = new byte[prelude.Length + preludeCRC.Length];
+                System.Buffer.BlockCopy(prelude,0,inputArray,0, prelude.Length);
+                System.Buffer.BlockCopy(preludeCRC,0,inputArray,prelude.Length, preludeCRC.Length);
+
+                // write real data to inputArray
+                Crc32Algorithm.ComputeAndWriteToEnd(inputArray); // last 4 bytes contains CRC
```
Praveenrajmani: @poornas We are actually overwriting the last 4 bytes (preludeCRC) with the calculated CRC in this function `ComputeAndWriteToEnd`

```
        /// <summary>
        /// Computes CRC-32 from input buffer and writes it after end of data (buffer should have 4 bytes reserved space for it). Can be used in conjunction with <see cref="IsValidWithCrcAtEnd(byte[],int,int)"/>
        /// </summary>
        /// <param name="input">Input buffer with data to be checksummed.</param>
        /// <param name="offset">Offset of the input data within the buffer.</param>
        /// <param name="length">Length of the input data in the buffer.</param>
        /// <returns>CRC-32 of the data in the buffer.</returns>
        public static uint ComputeAndWriteToEnd(byte[] input, int offset, int length)
        {
            if (length + 4 > input.Length)
                throw new ArgumentOutOfRangeException("length", "Length of data should be less than array length - 4 bytes of CRC data");
            var crc = Append(0, input, offset, length);
            var r = offset + length;
            input[r] = (byte)crc;
            input[r + 1] = (byte)(crc >> 8);
            input[r + 2] = (byte)(crc >> 16);
            input[r + 3] = (byte)(crc >> 24);
            return crc;
        }
 
        /// <summary>
        /// Computes CRC-32 from input buffer - 4 bytes and writes it as last 4 bytes of buffer. Can be used in conjunction with <see cref="IsValidWithCrcAtEnd(byte[])"/>
        /// </summary>
        /// <param name="input">Input buffer with data to be checksummed.</param>
        /// <returns>CRC-32 of the data in the buffer.</returns>
        public static uint ComputeAndWriteToEnd(byte[] input)
        {
            if (input.Length < 4)
                throw new ArgumentOutOfRangeException("input", "input array should be 4 bytes at least");
            return ComputeAndWriteToEnd(input, 0, input.Length - 4);
        }
```
Our `inputArray` here which is of 12bytes length has 8bytes(prelude)+4(preludeCRC) and this function `ComputeAndWriteToEnd` will calculate the CRC for (n - tailing 4) bytes and write the CRC to the last 4 bytes in inputArray. Which will overwrite the preludeCRC.

I think we should not copy the preludeCRC before validating. ie, We can do something like

```
var inputArray = new byte[prelude.Length + 4]; // 4 is the space for storing the calculated CRC
System.Buffer.BlockCopy(prelude,0,inputArray,0, prelude.Length); // Copy the prelude only
Crc32Algorithm.ComputeAndWriteToEnd(inputArray); 
// This will calculate the CRC for the prelude (first 8 bytes of the inputArray) and write the
// calculated CRC in the last 4 bytes of the inputArray
..
..
// We can then check if the last 4 bytes match the `preludeCRC` obtained from the input stream.
// (OR)
// We could calculate the crc by `var crc = Append(0, input, offset, length);` and then check if it 
// matches preludeCRC from the input stream.
```

Minio/ApiEndpoints/ObjectOperations.cs
```diff
@@ -182,6 +182,41 @@ public async Task GetObjectAsync(string bucketName, string objectName, string fi
...
+            var selectReqBytes = System.Text.Encoding.UTF8.GetBytes(opts.MarshalXML());
+
+            string resourcePath = "?select=&select-type=2";
+            var request = await this.CreateRequest(Method.POST, bucketName,
+                                                    objectName: objectName,
+                                                    resourcePath: resourcePath,
+                                                    headerMap: sseHeaders)
+                                    .ConfigureAwait(false);
+
+            request.AddParameter("application/xml", selectReqBytes, ParameterType.RequestBody);
```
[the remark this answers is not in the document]
poornas: [here](https://github.com/minio/minio-dotnet/blob/master/Minio/V4Authenticator.cs#L112) - before authentication.

Minio/DataModel/Select/SelectResponseStream.cs
poornas: good catch - fixed this

Minio/DataModel/Select/SelectResponseStream.cs
```diff
@@ -0,0 +1,239 @@
...
+                var n =  ReadFromStream(prelude);
+                numBytesRead += n;
+                n =  ReadFromStream(preludeCRC);
+                numBytesRead += n;
+                var inputArray = new byte[prelude.Length + preludeCRC.Length + 4];
+                System.Buffer.BlockCopy(prelude,0,inputArray,0, prelude.Length);
+                System.Buffer.BlockCopy(preludeCRC,0,inputArray,prelude.Length, preludeCRC.Length);
+
+                // write real data to inputArray
+                Crc32Algorithm.ComputeAndWriteToEnd(inputArray); // last 4 bytes contains CRC
```
Praveenrajmani: Here the `computeAndWriteToEnd` calculates the CRC for the first 8(prelude) + 4(preludeCRC) bytes and writes the CRC in the last 4 byte block. We should not included the preludeCRC in the CRC calculation.

let the inputArray be `var inputArray = new byte[prelude.Length + 4];` and let us copy the prelude alone to the inputArray and let `ComputeAndWriteToEnd` calculate the CRC for the first 8bytes (prelude) and write that in the last 4 byte block. We could then verify this calculated CRC in the last 4bytes with the preludeCRC that we got from the stream - (`n =  ReadFromStream(preludeCRC);`)

Minio/DataModel/Select/SelectResponseStream.cs
```diff
@@ -0,0 +1,239 @@
...
+
+                numBytesRead += num;
+                num = ReadFromStream(messageCRC);
+                // now verify message CRC
+                inputArray = new byte[totalLength + 4];
+                System.Buffer.BlockCopy(prelude,0,inputArray,0, prelude.Length);
+                System.Buffer.BlockCopy(preludeCRC,0,inputArray,prelude.Length, preludeCRC.Length);
+                System.Buffer.BlockCopy(headers,0,inputArray,prelude.Length+ preludeCRC.Length, headerLength);
+                System.Buffer.BlockCopy(payload,0,inputArray,prelude.Length+ preludeCRC.Length+headerLength, payloadLength);
+                System.Buffer.BlockCopy(messageCRC,0,inputArray, totalLength - 4,messageCRC.Length);
```
Praveenrajmani: We should exclude messageCRC for the crc calculation here as well

Minio/DataModel/Select/SelectResponseStream.cs
```diff
@@ -0,0 +1,239 @@
...
+                    {
+                        Console.WriteLine("payload|"+Encoding.UTF7.GetString(payload));
+                        StatsMessage stats = new StatsMessage();
+                        using (var stream = new MemoryStream(payload))
+                        stats = (StatsMessage)new XmlSerializer(typeof(StatsMessage)).Deserialize(stream);
+                        this.Stats = stats;
+                    }
+                    if (value.Equals("Records"))
+                    {
+                        this.Payload.Write(payload,0, payloadLength);
```
Praveenrajmani: If i understand correctly, `Payload` is the response writer that the caller/user will be listening to? @poornas

Minio/V4Authenticator.cs
```diff
@@ -383,22 +384,22 @@ internal string GetPresignCanonicalRequest(IRestClient client, IRestRequest requ
             // QUERY
             if (path.Length == 2)
             {
-                var parameterString = path[1];
-                var parameterList = parameterString.Split('&');
-                var sortedQueries = new SortedSet<string>();
-                foreach (string individualParameterString in parameterList)
+                var queryParams = path[1].Split('&').Select(p => p.Split('='))
```
Praveenrajmani: This change is not related to Select right?

Minio/DataModel/Select/SelectResponseStream.cs
poornas: fixed.

Minio/DataModel/Select/SelectResponseStream.cs
poornas: that's correct.

Minio/V4Authenticator.cs
poornas: Query parameters were not being canonicalized correctly - a preexisting bug exposed while implementing the select API. Hence this is a part of the PR.

Minio/DataModel/Select/SelectResponseStream.cs
poornas: fixed

poornas:
PR updated @Praveenrajmani & @krisis

Minio/DataModel/Select/SelectResponseStream.cs
```diff
@@ -0,0 +1,256 @@
...
+
+                numBytesRead += num;
+                num = ReadFromStream(messageCRC);
+                var messageCRCBytes = messageCRC.ToArray();
+                if (BitConverter.IsLittleEndian)
+                {
+                    Array.Reverse(messageCRCBytes); 
+                }
+                // now verify message CRC
+                inputArray = new byte[totalLength];
```
Praveenrajmani: Should we allocate 4 more bytes for the messageCRC check? like we did for the preludeCRC above?

Minio/DataModel/Select/SelectResponseStream.cs
```diff
@@ -0,0 +1,256 @@
...
+                System.Buffer.BlockCopy(prelude,0,inputArray,0, prelude.Length);
+                System.Buffer.BlockCopy(preludeCRC,0,inputArray,prelude.Length, preludeCRC.Length);
+                System.Buffer.BlockCopy(headers,0,inputArray,prelude.Length+ preludeCRC.Length, headerLength);
+                System.Buffer.BlockCopy(payload,0,inputArray,prelude.Length+ preludeCRC.Length+headerLength, payloadLength);
+
+                // write real data to inputArray
+                Crc32Algorithm.ComputeAndWriteToEnd(inputArray); // last 4 bytes contains CRC
```
Praveenrajmani: So, this overwrites the last four bytes of the payload. `inputArray = new byte[totalLength+4];` may fix this.

poornas: @Praveenrajmani , totalLength includes the entire 16 byte overhead including 4 for the messageCRC. The messageCRC is not copied into the inputArray, leaving 4 bytes for the CRC32 algorithm to store the computed value. Seems correct to me.

Issue #368, opened 2020-01-11
StatObjectAsync is called on any GetObjectAsync then result is parsed but ignored


Vfialkin:
Every call to GetObjectAsync results in a call to StatObjectAsync but we do don't return Stats data back.

(https://github.com/minio/minio-dotnet/blob/master/Minio/ApiEndpoints/ObjectOperations.cs)

**Problems:**
1) Its not obvious, even in your own examples you call to  StatObjectAsync before GetObjectAsync to verify that object exists but its not necessary
(https://docs.min.io/docs/dotnet-client-api-reference.html#getObject)

2) StatObjectAsync not just verifies if object exists but it also parses all returned headers, including metadata to create a StatsObject. And then we ignore it. That's a lot of string parsing and allocations for nothing, specially if you have many keys in metaData field.

3) There is no method to get Stream and StatsObject in one call, so I need to make two calls that actually will result in 3 calls and parsing same StatsObject twice.

**Suggestion:**
1) Create an internal override for StatObjectAsync that will just check but not return data. This will save resources on parsing headers when we don't need StatsObject but only proof of existence.

2) Create an override for GetObjectAsync that will return Task<StatsObject> instead of just Task

This way it will not break existing APIs.

In the next major version I recommend to rename StatObjectAsync to GetStatObjectAsync

kannappanr:
@Vfialkin Thanks for filing this issue. We will take a look and get back to you. Please feel free to send a PR

poornas:
Fixed by #369

PR #369, opened 2020-01-19
Fix for Issue: StatObjectAsync is called on any GetObjectAsync then r…

Vfialkin:
Fix for:
https://github.com/minio/minio-dotnet/issues/368#issuecomment-573566966

Split StatObjectAsync into two internal methods one to get a response, another to parse it to StatsObject.

Removed parsing part from methods where the result was not used.


poornas:
@Vfialkin , can you just remove the StatObjectAsync calls at [line 49](https://github.com/minio/minio-dotnet/blob/master/Minio/ApiEndpoints/ObjectOperations.cs#L49) and [line 88](https://github.com/minio/minio-dotnet/blob/master/Minio/ApiEndpoints/ObjectOperations.cs#L88)

the rest of the changes here are unnecessary

Praveenrajmani:
ping @Vfialkin

Issue #374, opened 2020-03-02
Not possible to overwrite the default Timeout


trnktms:
When setting up a `MinioClient` object I did not find the way to overwrite the default `RestClient` or `RestRequest` timeout setting.

PR #375, opened 2020-03-02
Extend MinioClient with WithTimeout

trnktms:
Fix for #374

### DIFF
--- a/Minio/MinioClient.cs
+++ b/Minio/MinioClient.cs
@@ -340,6 +340,17 @@ namespace Minio
         }
 
         /// <summary>
+        /// Uses the set timeout for all requests if this method is invoked on client object
+        /// </summary>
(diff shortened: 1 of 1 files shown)


Minio/MinioClient.cs
```diff
@@ -339,6 +339,17 @@ public MinioClient WithProxy(IWebProxy proxy)
             return this;
         }
 
+        /// <summary>
+        /// Uses the set timeout for all requests if this method is invoked on client object
+        /// </summary>
+        /// <param name="timeout">Timeout in miliseconds.</param>
```
krisis:
```suggestion
        /// <param name="timeout">Timeout in milliseconds.</param>
```

trnktms: Thank you! I applied your suggested change.

Issue #372, opened 2020-02-21
CopyObjectAsync is much slower than mc tool


ngbrown:
In `CalculateMultiPartSize`, which is triggered if the source is over 5 GiB, it appears the copy is made to maximize the number of parts (which is 10,000), bounded by the minimum part size (5 MiB).

https://github.com/minio/minio-dotnet/blob/b7e7adff75bfe63842063baa6b0f46b1054d0957/Minio/Helper/utils.cs#L235

On the other hand, the `mc` command line copies with much fewer parts.  It seems to optimize for a higher part size (500 MiB), or 100 times larger than the minimum part size.

The speed difference is significant.  A 7 GiB file copies between buckets on the commandline in 1 minute.  With the .net client, I canceled after 25 minutes, and it was only 56% done.

Is there any way to tune this?

kannappanr:
@ngbrown Thanks for filing this issue. As far as `copyobjectpart` is concerned, I agree with you since the data is already in the backend. We can use 5GB part size.

ngbrown:
The go client seems to do it quite a bit differently by getting the number of parts first:

https://github.com/minio/minio-go/blob/8e3928466d18a04caf35a76a0f4a41d65ab1c9fc/api-compose-object.go#L525

I did start a branch to use be able to use different minimum part sizes.  It's not tune-able like I suggested, but it is a lot faster.

https://github.com/ngbrown/minio-dotnet/commit/7f2acd121d22dd59e01b2295d4bde71c03ec5175

If you have a suggestion on which way to go, I can update my branch and start a pull request.

poornas:
@ngbrown, the reason mc is much faster than dotnet SDK here is because minimum part size used by mc is 128MiB vs the much lower 5MiB on dotnet SDK. We can make the part size configurable so that you can optionally bump it up if your network can handle it.

kannappanr:
@ngbrown, we can set the part size in the case of `CopyObjectPart` to `5GiB` and the `PutObjectPart` can still be `5MiB`. In another PR we can set the default to be `64MiB`, but also allow clients to override that value. Also, please use already defined constants for the numbers, `MaximumPartSize` and `MinimumPartSize`

ngbrown:
Regarding setting the part size to `5GiB`, I used `512 MiB` because parts completing appears to be the only way that there can be any progress reported on server copies.

[...]

PR #377, opened 2020-03-03
Make multipart sizes configurable and increase defaults

ngbrown:
Addresses issue #372 to speed up multipart uploads and inter-server transfers by making the client more configurable with higher defaults for multipart part sizes.

* Adds a constructor to `MinioClient` that takes a single `MinioClientConfig` object for future extensibility.
* Adds new option `DefaultServerTransferPartSize`, with a default of 512 MiB
* Removes use of `dynamic` and uses value tuple returns from `CalculateMultiPartSize()`
* Added additional tests for a configurable `minimumPartSize` in `CalculateMultiPartSize()`.
* Adds `MinioPutObjectOptions` object for `PutObjectAsync`, reflecting Go client.  Includes a new option `PartSize`, with a default of 64 MiB

### DIFF
--- a/Minio/MinioClientConfig.cs
+++ b/Minio/MinioClientConfig.cs
@@ -1,0 +1,60 @@
+﻿/*
+ * MinIO .NET Library for Amazon S3 Compatible Cloud Storage,
+ * (C) 2017, 2018, 2019, 2020 MinIO, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Minio
+{
+    /// <summary>
+    /// Configuration parameters for <see cref="MinioClient"/>
+    /// </summary>
+    public class MinioClientConfig
+    {
+        /// <summary>
+        /// Location of the server, supports HTTP and HTTPS
+        /// </summary>
+        public string Endpoint { get; set; }
+
+        /// <summary>
+        /// Connect to Cloud Storage with HTTPS
+        /// </summary>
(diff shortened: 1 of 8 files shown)


Minio/MinioClientConfig.cs
```diff
@@ -0,0 +1,40 @@
+﻿namespace Minio
```
harshavardhana: Add missing copyright header

Minio/MinioClientConfig.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        /// <summary>
+        /// Optional session token
+        /// </summary>
+        public string SessionToken { get; set; } = "";
+
+        /// <summary>
+        /// Change the default for multipart part sizes when uploading through client
+        /// </summary>
+        public long DefaultUploadPartSize { get; set; } = 64 * 1024L * 1024L;
```
harshavardhana: This is not a client level config, it should be part of PutObject input. That would provide the required flexibility per Upload operation.

ngbrown: It is intended to be a client level config, and probably one that doesn't have to be adjusted.    The behavior of `PutObjectAsync` changes when the object size goes above this limit.  Using the minimum part size is simply too small for most cases.

There are already three overloads for `PutObjectAsync`, and to add another parameter to them would result in at least one more.

Minio/MinioClientConfig.cs
ngbrown: Added copyright header.

Minio/MinioClientConfig.cs
harshavardhana: We should add one more in line with our other SDKs such as minio-go https://docs.min.io/docs/golang-client-api-reference.html#PutObject 

It is incorrect to have this configured for MinioClient constructor config, this should be configurable per PutObject() instead.

ngbrown: Alright.  I'll look more into what the minio-go client is doing.

ngbrown: I've move the extra parameters for `PutObjectAsync` into a new `MinioPutObjectOptions` and moved `DefaultUploadPartSize` option from `MinioClientConfig` to the new `MinioPutObjectOptions`.

harshavardhana:

ngbrown:
@harshavardhana I see that your new default for PUT in #531 is now 16 MB. That's still quite small and would take less than 0.1 seconds per part at 1000 Mbps. For server-to-server transactions, making this configurable would still be beneficial.

PR #382, opened 2020-03-12
Fix for issue: MakeBucketAsync returns deserialization exception if BucketName is null

AdamHodgson:
Throw ```InvalidBucketNameException``` from ```MakeBucketAsync``` when bucketName name is null.

### DIFF
--- a/Minio/ApiEndpoints/BucketOperations.cs
+++ b/Minio/ApiEndpoints/BucketOperations.cs
@@ -63,8 +63,14 @@ namespace Minio
         /// <param name="location">Region</param>
         /// <param name="cancellationToken">Optional cancellation token to cancel the operation</param>
         /// <returns> Task </returns>
+        /// <exception cref="InvalidBucketNameException">When bucketName is null</exception>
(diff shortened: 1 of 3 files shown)


Minio/ApiEndpoints/BucketOperations.cs
```diff
@@ -63,8 +63,14 @@ public async Task<ListAllMyBucketsResult> ListBucketsAsync(CancellationToken can
         /// <param name="location">Region</param>
         /// <param name="cancellationToken">Optional cancellation token to cancel the operation</param>
         /// <returns> Task </returns>
+        /// <exception cref="InvalidBucketNameException">When bucketName is null</exception>
         public async Task MakeBucketAsync(string bucketName, string location = "us-east-1", CancellationToken cancellationToken = default(CancellationToken))
         {
+            if (bucketName == null)
+            {
+                throw new InvalidBucketNameException(bucketName, "buckentName cannot be null");
```
krisis: typo:
```suggestion
                throw new InvalidBucketNameException(bucketName, "bucketName cannot be null");
```

PR #391, opened 2020-04-09
Fix error response parsing

### LINKED ISSUE #390: [Bug] BucketNotFound exception thrown on invalid file path
**Package version:** Minio v3.1.10

**Current behavior:**
```
var path = $"/dir/file.txt"; // leading '/'
await minioClient.StatObjectAsync("existingBucket", path)' // (1)
```
(1) throws BucketNotFoundException. Please note that the error reproduces with any api that calls the ParseError() method.

The reason is [here](https://github.com/minio/minio-dotnet/blob/master/Minio/MinioClient.cs#L438-L441). Values:
```
response.Request.Resource: "existingBucket//dir/file.txt"
objectName: ""
response: "StatusCode: NotFound, Content-Type: , Content-Length: 0)"
```

**Expected behavior:** the code should throw ObjectNotFoundException.

As a suggestion:
1. Please do not call `response.Request.Resource.Split()` five times in a row. It's excessive and leads to unintended GC allocations.
2. Use StringSplitOptions :)

poornas:
Fixes #390

### DIFF
--- a/Minio/MinioClient.cs
+++ b/Minio/MinioClient.cs
@@ -423,18 +423,19 @@ namespace Minio
                     }
 
                     errorResponse.Resource = response.Request.Resource;
+                    var resourceSplits = response.Request.Resource.Split('/');
 
                     if (HttpStatusCode.NotFound.Equals(response.StatusCode))
                     {
-                        int pathLength = response.Request.Resource.Split('/').Count();
+                        int pathLength = resourceSplits.Count();
                         bool isAWS = response.ResponseUri.Host.EndsWith("s3.amazonaws.com");
                         bool isVirtual = isAWS && !response.ResponseUri.Host.StartsWith("s3.amazonaws.com");
 
                         if (pathLength > 1)
                         {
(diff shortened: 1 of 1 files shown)


Minio/MinioClient.cs
```diff
@@ -423,18 +423,19 @@ internal static void ParseError(IRestResponse response)
...
                         bool isAWS = response.ResponseUri.Host.EndsWith("s3.amazonaws.com");
                         bool isVirtual = isAWS && !response.ResponseUri.Host.StartsWith("s3.amazonaws.com");
 
                         if (pathLength > 1)
                         {
                             errorResponse.Code = "NoSuchKey";
-                            var bucketName = response.Request.Resource.Split('/')[0];
-                            var objectName = response.Request.Resource.Split('/')[1];
+                            var bucketName = resourceSplits[0];
+                            var objectName = String.Join(",", resourceSplits.Skip(1));
```
Praveenrajmani, line 438: One minor nit @poornas - Shouldn't we join by `\` here?

PR #394, opened 2020-04-18
Improve region matching regex performance (fixes #393)

### LINKED ISSUE #393: GetRegionFromEndpoint introduces significant delay in requests
When a request is made to an endpoint with a long enough non-AWS hostname without specifying the region, the regex inside `GetRegionFromEndpoint` can take several seconds to complete. This slows down requests by quite a lot.

This happens because the regex contains two nested quantifiers (`{1,61}` and `*?`), which causes an exponential amount of backtracking on all endpoints that do not match the regex.

As an example, the regex will take about 25 seconds to run on my machine when given this endpoint: `aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.local`

I added the endpoints from the unit tests and an example endpoint to regex101 [here](https://regex101.com/r/8FGJD2/2/), which also warns about this:

>Catastrophic backtracking has been detected and the execution of your expression has been halted.

It would be nice if this regex could be simplified so it does not contain any nested quantifiers anymore. I would like to submit a PR for this, but I'm not sure if there are any other types of endpoints that need to be accounted for (besides the ones specified in the unit tests).

geluk:
In addition to simplifying the regex, I have also modified it to be compiled. This introduces a slight startup delay (measured to be around 4ms on my machine) but speeds up all subsequent evaluations. Endpoints under 200 characters in length consistently get evaluated in under 1ms during my tests.

Fixes #393.

### DIFF
--- a/Minio.Tests/RegionTest.cs
+++ b/Minio.Tests/RegionTest.cs
@@ -30,6 +30,8 @@ namespace Minio.Tests
         [DataRow("s3.us-west-1.amazonaws.com", "us-west-1")]
         [DataRow("mybucket-s3-us-west-1.amazonaws.com", "us-west-1")]
         [DataRow("wests3iss.s3-us-west-1.amazonaws.com", "us-west-1")]
+        [DataRow("test.s3-s3.bucket.s3-us-west-1.amazonaws.com", "us-west-1")]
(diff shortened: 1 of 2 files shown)


avanveelen:
Nice clean fix, this increased the performance of my application. Thanks for your work @geluk

Issue #395, opened 2020-04-21
[Suggestion] Typed exception for invalid object names


ig-sinicyn:
MinioClient v3.1.11

**Current behavior:**
requests with invalid object names throw InternalClientException, as example
```cs
var blobName = new string('A', 260);
await MinioClient.GetObjectAsync(
	bucketName,
	blobName,
	s => { ... });
```

**Expected behavior:**
Code should throw InvalidObjectNameException.

kannappanr:
@ig-sinicyn Please feel free to send a PR

ig-sinicyn:
Done, #398

poornas:
Fixed by #398

PR #398, opened 2020-04-22
Fix #396, #395: Use typed exceptions for minio client

### LINKED ISSUE #396: [Suggestion] Fixes for MinioException.
The MinioException has two public properties, `message` and `response`.

I'd suggest to use standard `Exception.Message` property. Client code may log only ex.Message. Having an empty message mages troubleshooting much harder. As it is for now, the `MinioException.Message` is null for some exceptions, BucketNotFoundException as example.

The `response` property should be named in PascalCase co conform c# naming guidelines. In addition, I believe it should be set for all exceptions where possible. Especially for the InternalClientException.
Currently there is no way to distinguish "server unavailable" error from "invalid object name" error. Both of them throw same InternalClientException with same properties.

ig-sinicyn:
* Rename MinioException properties: message => ServerMessage; response => ServerResponse.
* Split ParseError method into simpler and shorter methods
* Throw exceptions for invalid object name and connectivity errors.
* Add NegativeTest for newly added exceptions
* Pass IRestResponse to connectivity-related and unexpected-error exceptions. 
* Add SelectObjectContentException and UnexpectedMinioException exceptions. Code now use specific exceptions and do not throw base MinioException anymore.

### DIFF
--- a/Minio.Tests/NegativeTest.cs
+++ b/Minio.Tests/NegativeTest.cs
@@ -1,5 +1,6 @@
 ﻿/*
- * MinIO .NET Library for Amazon S3 Compatible Cloud Storage, (C) 2017 MinIO, Inc.
+ * MinIO .NET Library for Amazon S3 Compatible Cloud Storage,
+ * (C) 2017, 2018, 2019, 2020 MinIO, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/Minio.Tests/NegativeTest.cs
+++ b/Minio.Tests/NegativeTest.cs
@@ -1,0 +1,72 @@
+﻿/*
+ * MinIO .NET Library for Amazon S3 Compatible Cloud Storage, (C) 2017 MinIO, Inc.
+ *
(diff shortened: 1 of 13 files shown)


ig-sinicyn:
Up :)

Minio.Tests/NegativeTest.cs
```diff
@@ -0,0 +1,72 @@
+﻿/*
+ * MinIO .NET Library for Amazon S3 Compatible Cloud Storage, (C) 2017 MinIO, Inc.
```
Praveenrajmani, line 2: We have to fix the YEAR here. In other places as well.

ig-sinicyn: Ok, will do

ig-sinicyn: Done:)

poornas: @ig-sinicyn, the year needs to be fixed only on the files touched by your PR. Could you revert the changes to rest of the files?

ig-sinicyn: Done:)

PR #356, opened 2019-12-12
Prefix x-amz-meta- for custom metadata

poornas:
if not specified.

Fixes: #353

PR #409, opened 2020-05-05
Correct StatObjectAsync header retrieval

### LINKED ISSUE #408: Metadata values in StatObjectAsync are converted to lowercase.
This line was introduced in #356:
https://github.com/minio/minio-dotnet/blob/148406fd0a3918c9ed12c89a8c8f240650801ec7/Minio/ApiEndpoints/ObjectOperations.cs#L843

The metadata values are stored with correct casing in Minio, but the `StatObjectAsync` is converting them to lowercase on retrieval.

Are there real cases where the values would be prefixed with `x-amz-meta-`? I thought this was only for the header names (key).

I don't expect my metadata values to be converted to lowercase.  An example is a `Content-Disposition` header where filenames are case sensitive.

ngbrown:
This fixes #408.

Pull request #356 applied a case change to retrieved header values in addition to removing `x-amz-meta-` from values.  The author @kannappanr confirmed that this was not the intention.  The intention was to remove `x-amz-meta-` from the header name, not the value.

This pull request applies the fix in a backward compatible way in that it stores both the original header name and the trimmed header name into the returned metadata.

### DIFF
--- a/Minio/ApiEndpoints/ObjectOperations.cs
+++ b/Minio/ApiEndpoints/ObjectOperations.cs
@@ -25,7 +25,6 @@ using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Reactive.Linq;
-using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Xml.Linq;
@@ -36,9 +35,6 @@ namespace Minio
     public partial class MinioClient : IObjectOperations
     {
         private readonly List<string> supportedHeaders = new List<string> { "cache-control", "content-encoding", "content-type", "x-amz-acl", "content-disposition" };
-
-        public static readonly Regex valueReplaceRegex = new Regex(@"^x-amz-meta-",
-            RegexOptions.IgnoreCase | RegexOptions.CultureInvariant | RegexOptions.Compiled);
 
         /// <summary>
         /// Get an object. The object will be streamed to the callback given by the user.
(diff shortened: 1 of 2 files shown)


Minio/ApiEndpoints/ObjectOperations.cs
```diff
@@ -838,9 +837,14 @@ public async Task<ObjectStat> StatObjectAsync(string bucketName, string objectNa
...
                 }
-                else if (supportedHeaders.Contains(parameter.Name, StringComparer.OrdinalIgnoreCase) || parameter.Name.StartsWith("x-amz-meta-", StringComparison.OrdinalIgnoreCase))
+                else if (supportedHeaders.Contains(parameter.Name, StringComparer.OrdinalIgnoreCase))
                 {
-                    metaData[parameter.Name] = parameter.Value.ToString().ToLowerInvariant().Replace("x-amz-meta-", string.Empty);
+                    metaData[parameter.Name] = parameter.Value.ToString();
+                }
+                else if (parameter.Name.StartsWith("x-amz-meta-", StringComparison.OrdinalIgnoreCase))
+                {
+                    metaData[parameter.Name] = parameter.Value.ToString();
```
poornas, line 846: @ngbrown , this can be removed -SDK should be stripping off the x-amz-meta- prefix from headers.

ngbrown: The idea is that it wasn't doing this up until now, so not including the original header will be a breaking change.  In my application's case, I was both setting and retreiving the headers with `x-amz-meta-` already prefixed.

Praveenrajmani: I think it is still ok to break in this case. We cannot be preserving both. Like @poornas explained, SDKs should be always stripping the `x-amz-meta-` prefix.

IMO, Let this be a breaking change. Perhaps we can bump up the SDK version. cc @kannappanr

ngbrown: I've removed the compatibility line and the headers are now stripped of the `x-amz-meta-` prefix.

PR #422, opened 2020-06-15
Add assume role support (#341)

max-dankow:
- Add Assume Role support
- Add general positive tests

The main concern is that for testing Assume Roles we have to use non-root user. I've created one on play.min.io, but it doesn't seem like the right solution. So I ask for your help, if there is a better solution to the problem.


max-dankow:
Up
@poornas

ig-sinicyn:
Up

harshavardhana:

Issue #428, opened 2020-06-30
Error when try use minio with yandex s3: The request signature we calculated does not match the signature you provided.


sergei66666:
Hello.

At first:
1) minio 3.1.13
2) restsharp 106.11.4
3) Windows 10 x64
4) s3 compatible storage from [yandex](https://cloud.yandex.com/services/storage)

If I use library as is - getting error: The request signature we calculated does not match the signature you provided.

If use WinSCP (it can s3) - all works fine.

After all day of different attempts i found, that if inside V4Authenticator
replace `this.SetHostHeader(request, client.BaseUrl.Host + ":" + client.BaseUrl.Port);`
with `this.SetHostHeader(request, client.BaseUrl.Host);` then all works fine.

I see issue (but can't find it), where man pointinig that problem in restsharp, which remove port from header.
But than I open Wireshark and see that port is sended.

So, i have 2 questions:
1) all other services works fine when endpoint:port in host header?
If works - yandex has error.
2) I am newbie in s3 auth headers, but.. for what we should use port inside V4Authenticator, if restsharp decide which port use for connection? (I see it in stream from `GetObjectAsync`. `WithSSL()` - port 443, without - 80)

sergei66666:
I am sorry, it is Duplicate of #321
With one note: not only 80 port

chrisbecke:
Interesting. The issue here seems to be related specifically to the use of port 80, as port 80 is implicit when the scheme is http it is not necessary to always provide it.

The V4Authenticator code builds a canonical hostname using the port always, the code that actually generates the signature would seem to follow a different code path with a less rigorous canonicalisation of provided host name.

RFC2616 specifies that the default port is used when not specified in the host header, but does not go to the level of even saying the port SHOULD or SHOULD NOT be included when it is the default.

sergei66666:
On yandex s3 error happen on 80 and 443 port.
I assume, that their nginx configuration doesn't pass port from host header. I wrote them about it, may be they fix it.

sergei66666:
Just to note: yandex team fixed it

PR #434, opened 2020-07-01
Remove port from HostHeader

### LINKED ISSUE #321: SignatureDoesNotMatch when port is 80
I've discovered an issue with using the minio dotnet client when connected to a minio server over port 80.
The signature generated by the client sets host header using the following code: 
`SetHostHeader(request, client.BaseUrl.Host + ":" + client.BaseUrl.Port);`
This is V4Authenicator.cs: line 114
The issue is, the server doesn't seem to include the port number with host resulting in a signature mismatch.

I've verify buy running the dot-net samples code and removing port number from SetHostHeader. The signature mismatch issue goes away.

kannappanr:
Some object storages do not honor standard ports as part of hostheader

Fixes #321, #428

### DIFF
--- a/Minio/V4Authenticator.cs
+++ b/Minio/V4Authenticator.cs
@@ -114,7 +114,11 @@ namespace Minio
             DateTime signingDate = DateTime.UtcNow;
             this.SetContentMd5(request);
             this.SetContentSha256(request);
-            this.SetHostHeader(request, client.BaseUrl.Host + ":" + client.BaseUrl.Port);
+            if (client.BaseUrl.Port == 80 || client.BaseUrl.Port == 443) {
+                this.SetHostHeader(request, client.BaseUrl.Host);
+            } else {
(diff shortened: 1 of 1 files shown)


Minio/V4Authenticator.cs
```diff
@@ -114,7 +114,11 @@ public void Authenticate(IRestClient client, IRestRequest request)
             DateTime signingDate = DateTime.UtcNow;
             this.SetContentMd5(request);
             this.SetContentSha256(request);
-            this.SetHostHeader(request, client.BaseUrl.Host + ":" + client.BaseUrl.Port);
+            if (client.BaseUrl.Port == 80 || client.BaseUrl.Port == 443) {
+                this.SetHostHeader(request, client.BaseUrl.Host);
```
Praveenrajmani, line 118: @kannappanr Can we add a comment saying that default ports need not be included for the signature calculation?

PR #441, opened 2020-07-02
fix erroneous flag set

### LINKED ISSUE #440: Strange line in code (compare bool with null)
Hello.

Here https://github.com/minio/minio-dotnet/blob/master/Minio/V4Authenticator.cs#L565 is very strange line.

1) `var deleteParm = request.Parameters.Any(p => p.Name.Equals("delete",StringComparison.OrdinalIgnoreCase));`
`deleteParm` is bool

2) `isMultiDeleteRequest = !(deleteParm == null) || (deleteParm.Equals(null));`
here `deleteParm` compared to null.

And it is cause of warning "The result of the expression is always 'false' since a value of type 'bool' is never equal to 'null' of type 'bool?'"

I'am do not know how it really should be. For my opinion it is should `isMultiDeleteRequest = deleteParm`
What do you think?

poornas:
fixes #440

### DIFF
--- a/Minio/V4Authenticator.cs
+++ b/Minio/V4Authenticator.cs
@@ -561,8 +561,7 @@ namespace Minio
                 var isMultiDeleteRequest = false;
                 if (request.Method == Method.POST)
                 {
-                    var deleteParm = request.Parameters.Any(p => p.Name.Equals("delete",StringComparison.OrdinalIgnoreCase));
-                    isMultiDeleteRequest = !(deleteParm == null) || (deleteParm.Equals(null));
+                    isMultiDeleteRequest = request.Parameters.Any(p => p.Name.Equals("delete",StringComparison.OrdinalIgnoreCase));
                 }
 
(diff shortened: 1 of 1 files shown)


Minio/V4Authenticator.cs
```diff
@@ -562,7 +562,7 @@ private void SetContentMd5(IRestRequest request)
                 if (request.Method == Method.POST)
                 {
                     var deleteParm = request.Parameters.Any(p => p.Name.Equals("delete",StringComparison.OrdinalIgnoreCase));
-                    isMultiDeleteRequest = !(deleteParm == null) || (deleteParm.Equals(null));
+                    isMultiDeleteRequest = deleteParm;
```
kannappanr, line 565: it can just be 
```
isMultiDeleteRequest = request.Parameters.Any(p => p.Name.Equals("delete",StringComparison.OrdinalIgnoreCase));
```

PR #487, opened 2020-11-13
PresingedPutObject API. Args & Functional test.

BigUstad:


Minio/DataModel/ObjectOperationsArgs.cs
```diff
@@ -86,4 +88,29 @@ public SelectObjectContentArgs WithRequestProgress(RequestProgress requestProgre
...
+        {
+            this.RequestMethod = Method.PUT;
+        }
+
+        internal new void Validate()
+        {
+            base.Validate();
+            if (!utils.IsValidExpiry(this.Expiry))
+            {
+                throw new InvalidExpiryRangeException("Expiry range should be between 1 and " + Constants.DefaultExpiryTime.ToString());
```
krisis, line 106: nit: shouldn't we say "... should be between 1s ..." since the unit of time is different for the lower and upper bound for expiry?

krisis (commented):
Overall changes look good to me.

Minio/DataModel/ObjectOperationsArgs.cs
BigUstad: Message updated @krisis . Tx. PTAL.

Minio.Functional.Tests/FunctionalTest.cs
```diff
@@ -939,7 +939,7 @@ internal async static Task PutObject_Task(MinioClient minio, string bucketName,
 
                 await minio.PutObjectAsync(bucketName,
                                             objectName,
-                                            filestream,
+                                                filestream,
```
sinhaashish, line 942: extra space

sinhaashish (approved):
just a minor comment . otherwise LGTM

krisis (approved):

Minio.Functional.Tests/FunctionalTest.cs
BigUstad: Tx for pointing out. Correcting it.

PR #506, opened 2020-12-21
Handle github action failure dependency installation.

BigUstad:


BigUstad:
The command added needs to be changed.

PR #510, opened 2021-01-25
Moving Tagging related DataModel files to Tag folder

BigUstad:


BigUstad:
Will need namespace change as well.

PR #511, opened 2021-01-25
Retention DataModel files moved into folder Retention

BigUstad:


BigUstad:
Will do it in a new PR

PR #512, opened 2021-01-25
Object Lock DataModel files moved to folder ObjectLock

BigUstad:


BigUstad:
Will need to put Lock/Retention in ObjectLock folder. Needs a new PR.

PR #527, opened 2021-04-13
fix: ListObjectsAsync error for empty bucket/prefix

### LINKED ISSUE #488: MinioClient.ListObjectsAsync() throws an error when the bucket is empty
As the title says, simple code:
```
            try
            {
                await this.client.ListObjectsAsync(bucketName);
                return true;
            }
            catch
            {
                return false;
            }
```
always returns false, when the requested bucket does not contain any objects (e.g. new bucket).

The caught exception is:
```
System.InvalidOperationException: Sequence contains no elements.
   at System.Reactive.Subjects.AsyncSubject`1.GetResult() in D:\a\1\s\Rx.NET\Source\src\System.Reactive\Subjects\AsyncSubject.cs:line 361
```

BigUstad:
Fixes #488

### DIFF
--- a/Minio.Examples/Cases/ListObjects.cs
+++ b/Minio.Examples/Cases/ListObjects.cs
@@ -25,7 +25,8 @@ namespace Minio.Examples.Cases
         public static void Run(MinioClient minio,
                                      string bucketName = "my-bucket-name",
                                      string prefix = null,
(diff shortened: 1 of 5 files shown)


kannappanr (approved):
LGTM

poornas (approved):
LGTM

PR #533, opened 2021-04-22
Func tests gateway, NotImplemented/Exception handling

BigUstad:
Signed-off-by: Priyank Raj <priyank.raj@gatech.edu>


BigUstad:
#535 covers it.

PR #545, opened 2021-05-25
Runs all MakeBucketAsync tests regardless of the platform

ebozduman:
Some `MakeBucketAsync` test cases were running only if the platform is AWS by checking if the environment variable, `"AWS_ENDPOINT"`, is `null` or not.
However, there is no good reason to have this distinction at this point. All existing test cases apply to both platforms.
So, this PR makes all `MakeBucketAsync` test cases run against both AWS and MinIO servers.


BigUstad (approved):
LGTM

harshavardhana (approved):

PR #546, opened 2021-05-25
User choice for ListObjects(V1)/ListObjectsV2

BigUstad:


Minio/DataModel/BucketOperationsArgs.cs
```diff
@@ -157,24 +181,41 @@ public GetObjectListArgs WithContinuationToken(string token)
...
             request.AddQueryParameter("delimiter",this.Delimiter);
-            request.AddQueryParameter("prefix",this.Prefix);
+            if (!string.IsNullOrWhiteSpace(this.Prefix))
+            {
+                request.AddQueryParameter("prefix",this.Prefix);
+            }
             request.AddQueryParameter("max-keys", "1000");
             request.AddQueryParameter("encoding-type","url");
-            if (!string.IsNullOrWhiteSpace(this.KeyMarker))
+            if (!string.IsNullOrWhiteSpace(this.KeyMarker) && this.UseV2)
```
kannappanr, line 207: useV2 is not important here, versions is important here

BigUstad: Changed. PTAL.

Minio/DataModel/BucketOperationsArgs.cs
```diff
@@ -157,28 +181,45 @@ public GetObjectListArgs WithContinuationToken(string token)
...
-            if (!string.IsNullOrWhiteSpace(this.VersionIdMarker))
+            if (!string.IsNullOrWhiteSpace(this.VersionIdMarker) && this.Versions)
             {
                 request.AddQueryParameter("version-id-marker",this.VersionIdMarker);
             }
+            if (!string.IsNullOrWhiteSpace(this.Marker))
+            {
+                request.AddQueryParameter((this.UseV2)?"start-after":"marker",this.Marker);
+            }
+            if (!this.UseV2)
```
kannappanr, line 219: isn't this check redundant? as it is already checked in the previous line?

BigUstad: Correction made.
We need an 'empty' marker if 'max-keys' is defined and the result is marked 'IsTruncated'. Else in the response <Marker> field isn't filled leading to an erroneous request-response situation.

BigUstad: * '<Marker>' in the response field isn't filled

kannappanr (approved):
LGTM

PR #552, opened 2021-07-05
feat: Add a method to obtain objects asynchronously.

netcore-jroger:


harshavardhana:

netcore-jroger:
when will be released?

harshavardhana:
> when will be released?

Soon @netcore-jroger we have few more things to take care of.

ebozduman:
@netcore-jroger,

You might already know about it, but I want to make sure we do have the `GetObjectListAsync` method.
Would it work for you in your scenario?

netcore-jroger:
no, I don't need the GetObjectListAsync method. @ebozduman

PR #566, opened 2021-12-13
Resolve MakeBucketAsync failure when specifying a region

realZhangChi:


realZhangChi:
Close because I see that the master branch has solved the problem.

PR #580, opened 2022-02-11
HttpClient injectable through MinioClient class consctructor

ebozduman:
This PR let's end user to have an option to define HttpClient and inject it through `.WithHttpServer()` method.


ebozduman:
Sorry for the confusion. Pulling it back...

Issue #411, opened 2020-05-11
Too many TCP connections established


yilezhu:
When I conduct the concurrent test of Minio, I find that if I upload files concurrently, a lot of TCP connections will be established and will not be released。
My test code is as follows:
`
static void Main(string[] args)
        {

            try
            {
                Run().Wait();
            }
            catch (Exception ex)
            {
                Console.WriteLine(ex.Message);
            }
            Console.ReadLine();
        }

        private static async Task Run()
        {
            // Make a new bucket called mymusic.
            var bucketName = "testsize"; //<==== change this
            var location = "us-east-1";
            var filePath = @"C:\Users\admin\Desktop\c143a9bc-be71-4475-9981-3c0d8f60d7f5";
            var contentType = "application/octet-stream";
            var endpoint = "192.168.1.14:9000";
            var accessKey = "minioadmin";
            var secretKey = "minioadmin";
            try
            {
                //FileStream fileStream = new FileStream(filePath, FileMode.Open, FileAccess.Read);
              
                var minio = new MinioClient(endpoint, accessKey, secretKey);

                bool found = await minio.BucketExistsAsync(bucketName);
                if (!found)
                {
                    await minio.MakeBucketAsync(bucketName, location);
                }

                Console.WriteLine("10万个文件上传开始~~~~" + DateTime.Now);
                Stopwatch stopwatch = Stopwatch.StartNew();
                await Task.WhenAll(Enumerable.Range(0, 100_000)
                .Select(_ => Task.Run(async () =>
                {
                    //fileStream.Position = 0;
                    await minio.PutObjectAsync(bucketName, Guid.NewGuid().ToString("N"), filePath, contentType);
                }))).ConfigureAwait(false);
                stopwatch.Stop();

                Console.WriteLine("10万个文件上传结束~~~~" + DateTime.Now);
                Console.WriteLine("总共用时" + stopwatch.ElapsedMilliseconds / 1000 / 60 + "分钟");
                //fileStream.Dispose();
            }
            catch (Exception e)
            {
                Console.WriteLine(e);
            }
        }
`
At this time, I run the test code and view the TCP link through the tcpview tool。
Found that the TCP connection has been established and has not been released。At this time, the CPU of my local computer has gone up to 100%。
![image](https://user-images.githubusercontent.com/18655233/81523981-3015fb80-9382-11ea-90f0-05d66298f78a.png)

[...]

PR #582, opened 2022-02-16
HttpClient instance as a property of MinioClient class

ebozduman:
This PR let's end user to have an option to create his/her own HttpClient instance and pass it into MinioClient constructor as an argument. 
The new method is `.WithHttpServer()` method.
HttpClient's socket leakage (#411 ) and DNS issues are also addressed.

### DIFF
--- a/Minio/MinioClient.cs
+++ b/Minio/MinioClient.cs
@@ -30,6 +30,8 @@ using Minio.DataModel;
 using Minio.DataModel.Tracing;
 using Minio.Exceptions;
 using Minio.Helper;
+using Microsoft.Extensions.DependencyInjection;
+
 
 namespace Minio
 {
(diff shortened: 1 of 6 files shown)


Minio/MinioClient.cs
```diff
@@ -467,6 +470,17 @@ public MinioClient WithRetryPolicy(RetryPolicyHandlingDelegate retryPolicyHandle
...
 
+        /// <summary>
+        /// Allows end user to define the Http server and pass it as a parameter
+        /// </summary>
+        /// <param name="httpClient"> Instance of HttpClient</param>
+        /// <returns></returns>
+        public MinioClient WithHttpClient(HttpClient httpClient)
+        {
+            this.HTTPClient = httpClient;
+            return this;
```
poornas, line 481: validate if HttpClient is null

poornas (changes requested):

Minio/MinioClient.cs
ebozduman: Done.

PR #586, opened 2022-03-02
Clean up unsupported sdks and add support for net6.0 sdk

ebozduman:
Microsoft does not support some of the .NET SDKs, Minio SDK was supporting, netstandard2.0 and net4.6, This PR removes these SDKs from the project target frameworks and adds the following versions currently supported at this point:
net6.0, net5.0 (and net3.1, but we did not support this version before). 
net7.0 is available at this point, but it is in Preview stage.
This information is valid for all 3 platforms, Windows, Linux and macOS.
Reference: https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core#cadence

### DIFF
--- a/Minio/Minio.csproj
+++ b/Minio/Minio.csproj
@@ -4,7 +4,7 @@
     <LangVersion>latest</LangVersion>
     <AssemblyName>Minio</AssemblyName>
     <RootNamespace>Minio</RootNamespace>
-    <TargetFrameworks>netstandard2.0;net46;net5.0;</TargetFrameworks>
+    <TargetFrameworks>net5.0; net6.0</TargetFrameworks>
     <DebugType>embedded</DebugType>
     <GeneratePackageOnBuild>False</GeneratePackageOnBuild>
     <AssemblyOriginatorKeyFile>..\Minio.snk</AssemblyOriginatorKeyFile>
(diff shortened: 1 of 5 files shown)


ebozduman:
Looking into the build failures...

Minio/Minio.csproj
```diff
@@ -4,7 +4,7 @@
     <LangVersion>latest</LangVersion>
     <AssemblyName>Minio</AssemblyName>
     <RootNamespace>Minio</RootNamespace>
-    <TargetFrameworks>netstandard2.0;net46;net5.0;</TargetFrameworks>
+    <TargetFrameworks>net5.0; net6.0</TargetFrameworks>
```
ramonesz297, line 7: Why netstandard2.0 has been removed? All legacy applications will not be able to get the latest fixes. There are a lot of apps built with netstandard2.0 and support that very important support them as well. You can take a look at my [repo](https://github.com/ramonesz297/minio-client) and some performance improvements for net5 and net6 can be used here without removing netstandard2.0 support.

ramonesz297 (changes requested):

Minio/Minio.csproj
harshavardhana: netstandard is not supported anymore by MSFT https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core#cadence

ramonesz297: @harshavardhana  Are you going to completely remove net4.x support?

ramonesz297:
@ebozduman Are you going to completely remove net4.x support?

harshavardhana:
> @ebozduman Are you going to completely remove net4.x support?

Yes we removing everything that is eol'ed by MSFT.

Minio/Minio.csproj
harshavardhana: Yes

ebozduman:
@ramonesz297 

.NET Framework 4.5.2, 4.6, and 4.6.1 will reach end of support* on April 26, 2022. It is only a month+ away.
So, as @harshavardhana said, we won't be supporting it.

ramonesz297:
@ebozduman What about net4.7 or net4.8? Microsoft still supporting these versions… 
See reference: https://docs.microsoft.com/en-us/lifecycle/products/microsoft-net-framework

Also, Microsoft committed, that all new versions (including net5 and net6) will continue using netsrandard2.0 specifications. 

IMHO, MinioClient does not use any net5 or net6 features and this project can target only netstandatd2.0 without any changes and risks.

ebozduman:
@ramonesz297 

Thank you for pointing out the still supported 4.[78] releases and actually 4.6.2.(at the minimum) too. I also noticed it and brought up with the team, so we'll add support for them.

Thank you also for bringing up the 2nd point about Minio staying with old releases rather than moving together with officially supported .NET releases only. 
I need to discuss it with the team. I am not sure how much support it'll get, but my understanding is that moving with supported releases is not all about Minio using new features of new releases. I'll communicate the response here.

harshavardhana:
> @ramonesz297 
> 
> Thank you for pointing out the still supported 4.[78] releases and actually 4.6.2.(at the minimum) too. I also noticed it and brought up with the team, so we'll add support for them.
> 
> Thank you also for bringing up the 2nd point about Minio staying with old releases rather than moving together with officially supported .NET releases only. 
> I need to discuss it with the team. I am not sure how much support it'll get, but my understanding is that moving with supported releases is not all about Minio using new features of new releases. I'll communicate the response here.

Bottom line is we will support whatever is currently supported and not EOLed by MSFT.

harshavardhana:
Closing as this is merged in PR #597

PR #596, opened 2022-03-12
Fix HttpClient connection limit issue

### LINKED ISSUE #595: HttpClient connection limit issue
There is a bug with the connection limit in the current implementation.
This issue can be easily reproduced with _console application on full .NET Framework 4.8_:
```csharp
var minio = new MinioClient().WithEndpoint("localhost:9000").WithCredentials("login", "pass").Build();

var statObjectArgs = new StatObjectArgs().WithBucket("bucket").WithObject("file.txt");
var stat = await minio.StatObjectAsync(statObjectArgs);
var stat2 = await minio.StatObjectAsync(statObjectArgs);

// This request throwns a timeout error because there are no available connections
var stat3 = await minio.StatObjectAsync(statObjectArgs);
```

_Explanation_

The MinIO client uses HttpClient for requests with option HttpCompletionOption.ResponseHeadersRead
https://github.com/minio/minio-dotnet/blob/9a3f51d0f984de9133fd6cb981dd9536136a36fe/Minio/MinioClient.cs#L564-L566
The responses are not disposed, in this case, the connection is not closed.
[A console application on .NET FW 4.8 connection limit is equal to 2.](https://docs.microsoft.com/en-us/dotnet/api/system.net.servicepointmanager.defaultconnectionlimit?view=netframework-4.8)
> The maximum number of concurrent connections allowed by a [ServicePoint](https://docs.microsoft.com/en-us/dotnet/api/system.net.servicepoint?view=netframework-4.8) object. The default connection limit is 10 for ASP.NET hosted applications and 2 for all others.

That's why the only first two calls in the example were completed successfully, but the third one was failed.
You can read more about the issue here https://github.com/dotnet/runtime/issues/30304#issuecomment-513240331

olsh:
The PR adds explicit disposing for ResponseResult class.
Fixes #595

### DIFF
--- a/Minio/ApiEndpoints/BucketOperations.cs
+++ b/Minio/ApiEndpoints/BucketOperations.cs
@@ -47,7 +47,7 @@ namespace Minio
             CancellationToken cancellationToken = default(CancellationToken))
         {
             var requestMessageBuilder = await this.CreateRequest(HttpMethod.Get).ConfigureAwait(false);
-            var response = await this.ExecuteTaskAsync(this.NoErrorHandlers, requestMessageBuilder, cancellationToken)
+            using var response = await this.ExecuteTaskAsync(this.NoErrorHandlers, requestMessageBuilder, cancellationToken)
                 .ConfigureAwait(false);
 
             ListAllMyBucketsResult bucketList = new ListAllMyBucketsResult();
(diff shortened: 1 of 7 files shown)


harshavardhana (approved):

harshavardhana:
PTAL build failures @olsh

olsh:
@harshavardhana the tests passed on the second run.
Probably it was a temporary issue with infrastructure?

PR #609, opened 2022-03-26
Fixes the default expiry as 3600 seconds

ebozduman:
The default expiry date was supposed to be 3600 seconds. This PR fixes that.


kannappanr (approved):
LGTM

PR #615, opened 2022-04-12
AssumeRole feature support

ebozduman:
AssumeRole feature implementation


ebozduman:
Looking into the build test failure. I'll also re-base.

Minio/V4Authenticator.cs
```diff
@@ -110,35 +111,51 @@ public string Authenticate(HttpRequestMessageBuilder requestBuilder)
...
+            string canonicalRequest = this.GetCanonicalRequest(requestBuilder, headersToSign, assumeRole);
+            byte[] canonicalRequestBytes;
+            if (assumeRole)
+            {
+                canonicalRequestBytes = Encoding.ASCII.GetBytes(canonicalRequest);
+            }
+            else
+            {
+                canonicalRequestBytes = Encoding.UTF8.GetBytes(canonicalRequest);
+            }
```
harshavardhana, line 127: this doesn't look correct @ebozduman - why do you need UTF8 v/s ASCII ?

Minio/V4Authenticator.cs
```diff
@@ -166,25 +184,52 @@ private string GetSignedHeaders(SortedDictionary<string, string> headersToSign)
...
+        /// </summary>
+        /// <param name="assumeRole">boolean; if true role credentials, otherwise IAM user</param>
+        /// <returns>returns the kind of service as a string</returns>
+        private string getServiceKind(bool assumeRole)
+        {
+            var serviceKind = "s3";
+            if (assumeRole) serviceKind = "sts";
+            return serviceKind;
+        }
+
```
harshavardhana, line 198: This is also an incorrect API `assumeRole` is a single API it should be called `getService(bool sts)`

harshavardhana (changes requested):

Minio/V4Authenticator.cs
ebozduman: @harshavardhana 

I agree. This is where I actually had a "To be discussed during review" comment, but during the debug/comment cleanup before my final push, it looks like I accidentally wiped it out. 

The original code was using UTF8 encoding. The AssumeRole test I used during the development kept failing at this point. I believe I checked on the server side and I thought the problem is the encoding type and when I tested with ASCII encoding, it worked. It did not make sense, so I decided to put a comment here so it can be discussed during the code review. 

What I will do is to remove the ASCII encoding and run the test to see if I can reproduce the issue.
Please let me know if you have something else to recommend.

Minio/V4Authenticator.cs
ebozduman: OK. I'll modify it.

harshavardhana (approved):

Issue #604, opened 2022-03-16
Full .NET framework support


olsh:
The latest MinIO version 4.0.1 doesn't support the full .NET framework and some other platforms. Now it is targeted only .NET 5, 6
I believe that the support was removed in this PR #597 

It would be nice to add `netstandard2.0` target because .NET full framework is still supported and used in many projects.
https://docs.microsoft.com/en-us/lifecycle/products/microsoft-net-framework
https://docs.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-0#select-net-standard-version

harshavardhana:
@olsh we may consider this in the future thanks for opening the issue.

ramonesz297:
@olsh You can take a look at this [repo](https://github.com/ramonesz297/minio-client) as an alternative. It supports netstandard2.0.

PR #627, opened 2022-05-09
Support netstandard 2.0 as package target

abramovandrey:
Currently it is not possible to use Minio client with .NET Framework applications

### DIFF
--- a/Minio/Minio.csproj
+++ b/Minio/Minio.csproj
@@ -3,7 +3,7 @@
     <LangVersion>latest</LangVersion>
     <AssemblyName>Minio</AssemblyName>
     <RootNamespace>Minio</RootNamespace>
-    <TargetFrameworks>net5.0;net6.0;</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net5.0;net6.0;</TargetFrameworks>
     <DebugType>embedded</DebugType>
     <GeneratePackageOnBuild>False</GeneratePackageOnBuild>
     <AssemblyOriginatorKeyFile>..\Minio.snk</AssemblyOriginatorKeyFile>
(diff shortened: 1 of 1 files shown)


harshavardhana:
Only plan is to support #626 - what is the need here?

abramovandrey:
https://github.com/minio/minio-dotnet/pull/626/commits/f729a9358daff73153675bf3fa9e118ac4fddd4a targets specific net48 and has compilation errors.
Supporting netstandard2.0 will target large amount of frameworks starting from 4.6.1 up to net6.0
That's why separate MR was created

abramovandrey:
> > Is there a ticket and\or MR to fix 5 failing tests in master branch?
> 
> you need to fix this in this PR - you are bringing in the netstandard2.0 change. run the tests locally and fix them.

They seem to be failing without netstandard as well. Master branch without netastandard changes has 5 failing tests. Is it possible to see fresh build of master branch somewhere on github?

harshavardhana:
This is the PR for master that builds fine https://github.com/minio/minio-dotnet/pull/624

The build failure is because of your changes

ebozduman:
@harshavardhana , @abramovandrey

I also saw the similar/same Minio.Tests failures.

I found a couple of issues:

1. It is @abramovandrey, it is your fix: missing `"Microsoft.CSharp"` package in Minio/Minio.csproj file,

2. `"netstandard2.0"` is EOL already. So, we need to go with `"netstandard2.1"` and we have to have TFM `net6.0` also because it brings more runtime and language features, or APIs, some of which we depend on. I also needed to upgrade TestAdapter, TestFramework and Test.Sdk packages as the previous TestAdapter together with `"netstandard2.1"` (or with 2.0) were hitting a dotnet bug which was causing the failure on Linux platform. 

Both of the above issues are addressed in PR #626 which was created for issue #604. 

So, #626 has fix both missing targets `"net4.8"` and `"netstandard2.1"`.
I suggest to address the missing targets in the same PR and have a reference to this PR in issue #604 and closed.

abramovandrey:
https://github.com/minio/minio-dotnet/pull/627#issuecomment-1121426838 nope, it's not my changes, it seems to be test stand issue.

@ebozduman ok, forget it :) I'll just use AWSSDK.S3 instead to save time.

Not everyone are writing dotnet applications 'from scratch'. Nestandard 2.0 is a kind of 'bridge' for legacy applications to move from .NET Framework to .NET using libraries with common logic written using netstandard 2.0.
This logic may include S3 storage operations as well.

harshavardhana:
Relevant changes are in already in #626

PR #631, opened 2022-05-11
fix:MinioClient constructor httpClient is null

JiashuoLin:


harshavardhana:
This PR is wrong, please use builder pattern.

PR #641, opened 2022-05-19
Reuse HttpClient and fix the wrong TLS and Proxy settings during MinioClient creation

ebozduman:
Socket/Memory leak issue was addressed before, but there was a case where we were not using HttpHandler when MinioClient is created with no args.
Also adds a functional test for socket/memory leak issue.

### DIFF
--- a/Minio.Functional.Tests/Program.cs
+++ b/Minio.Functional.Tests/Program.cs
@@ -17,7 +17,6 @@
 
 using System;
 using System.Net;
-using System.Net.Http;
 
 namespace Minio.Functional.Tests;
 
@@ -51,25 +50,17 @@ internal class Program
 
         MinioClient minioClient = null;
 
-        var clientHandler = new HttpClientHandler();
-        clientHandler.ServerCertificateCustomValidationCallback =
-            (sender, cert, chain, sslPolicyErrors) => { return true; };
-        clientHandler.UseProxy = false;
-        var httpClient = new HttpClient(clientHandler);
-
         if (enableHttps == "1")
             // WithSSL() enables SSL support in MinIO client
             minioClient = new MinioClient()
                 .WithSSL()
                 .WithCredentials(accessKey, secretKey)
                 .WithEndpoint(endPoint)
-                .WithHttpClient(httpClient)
                 .Build();
         else
             minioClient = new MinioClient()
                 .WithCredentials(accessKey, secretKey)
                 .WithEndpoint(endPoint)
-                .WithHttpClient(httpClient)
                 .Build();
 
         // Assign parameters before starting the test
--- a/Minio.Functional.Tests/Program.cs
+++ b/Minio.Functional.Tests/Program.cs
@@ -66,9 +66,10 @@ internal class Program
                 .WithHttpClient(httpClient)
                 .Build();
         else
-            minioClient = new MinioClient(httpClient)
+            minioClient = new MinioClient()
                 .WithCredentials(accessKey, secretKey)
                 .WithEndpoint(endPoint)
+                .WithHttpClient(httpClient)
                 .Build();
 
         // Assign parameters before starting the test
--- a/Minio.Functional.Tests/Program.cs
+++ b/Minio.Functional.Tests/Program.cs
@@ -68,7 +68,7 @@ internal class Program
                 .WithHttpClient(httpClient)
                 .Build();
         else
-            minioClient = new MinioClient()
+            minioClient = new MinioClient(httpClient)
                 .WithCredentials(accessKey, secretKey)
                 .WithEndpoint(endPoint)
                 .Build();
--- a/Minio.Functional.Tests/Program.cs
+++ b/Minio.Functional.Tests/Program.cs
@@ -52,10 +52,8 @@ internal class Program
         MinioClient minioClient = null;
 
         var clientHandler = new HttpClientHandler();
-        clientHandler.ServerCertificateCustomValidationCallback = (sender, cert, chain, sslPolicyErrors) =>
-        {
-            return true;
-        };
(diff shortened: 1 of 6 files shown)


harshavardhana:
> Running `dotnet regitlint --fail-on-diff` for the build issue

You need fix, your formatting is all wrong for a .Net project source. Also you can't remove TLS skip for verification. Or you would have to trust the certs in CI/CD

ebozduman:
OK. I'll add TLS back.

About formatting:
I use `VisualStudio` as my editor and I've disabled `format on save` setting, but it still forces format changes. I'll check it in again. I'll also make sure I manually run `dotnet regitlint` before pushing the changes going forward.

Minio.Functional.Tests/Program.cs
```diff
@@ -68,7 +66,7 @@ public static void Main(string[] args)
                 .WithHttpClient(httpClient)
                 .Build();
         else
-            minioClient = new MinioClient()
+            minioClient = new MinioClient(httpClient)
                 .WithCredentials(accessKey, secretKey)
```
harshavardhana, line 70: You should use WithHttpClient to follow builder pattern.

Minio/MinioClient.cs
```diff
@@ -437,6 +441,11 @@ public MinioClient WithRetryPolicy(RetryPolicyHandlingDelegate retryPolicyHandle
     /// <returns></returns>
     public MinioClient WithHttpClient(HttpClient httpClient)
     {
+        if (HTTPClient != null)
```
harshavardhana, line 444: This function should throw an exception for httpClient bring null which is the input as InvalidArgumentException

harshavardhana: Also this If condition is not needed unless we kind of support all kinds of constructors, we should perhaps remove old style constructors (deprecated ones), move to 5.0.0 and drop all of their support.

harshavardhana (approved):

harshavardhana (changes requested):
Some reviews

Minio.Functional.Tests/Program.cs
ebozduman: Ok

ebozduman: Done.

ebozduman:
Implemented review comments and removed the socket/memory test per team decision.

Marked the parameterized construction method, `MinioClient(HttpClient httpClient)`, obsolete. If user has this method used in his/her code, then s/he will see a message similar to the following warning message:
```
.../Minio.Functional.Tests/Program.cs(69,27): warning CS0618: 'MinioClient.MinioClient(HttpClient)' is obsolete: 
'Use MinioClient() and Builder method .WithHttpClient(httpClient)' [.../Minio.Functional.Tests/Minio.Functional.Tests.csproj]
```

Minio/MinioClient.cs
ebozduman: Done.

Minio/MinioClient.cs
```diff
@@ -87,13 +87,18 @@ public MinioClient()
         Region = "";
         SessionToken = "";
         Provider = null;
-        HTTPClient = new HttpClient();
+        var clientHandler = new HttpClientHandler();
+        clientHandler.ServerCertificateCustomValidationCallback =
+            (sender, cert, chain, sslPolicyErrors) => { return true; };
+        clientHandler.UseProxy = false;
```
harshavardhana, line 93: You can't do this either - we should honor local system proxies.

Minio/MinioClient.cs
```diff
@@ -87,13 +87,18 @@ public MinioClient()
         Region = "";
         SessionToken = "";
         Provider = null;
-        HTTPClient = new HttpClient();
+        var clientHandler = new HttpClientHandler();
+        clientHandler.ServerCertificateCustomValidationCallback =
+            (sender, cert, chain, sslPolicyErrors) => { return true; };
```
harshavardhana, line 92: if you do this then SSL verification will be turned-off in the SDK we cannot have an insecure TLS enforcing HTTP client, we need to make sure by default we ask for trust and security. 

I am not why this code is needed.

harshavardhana (changes requested):

Minio/MinioClient.cs
ebozduman: OK.
I'll remove it.

Minio/MinioClient.cs
ebozduman: I'll remove it.

ebozduman: Removed

harshavardhana:
I am not sure what is the bug fix here? @ebozduman - can you please explain?

ebozduman:
Sorry for the confusion.
Things changed in time and the title and what has changed diverged.

TLDR: This PR will make the last 3 test cases reuse the existing HttpClient instance and fix the wrong TLS and Proxy settings when creating an instance of MinioClient.

Some more explanations:
After the original socket leakage problem is addressed (too many HttpClient instances were created instead of reusing the existing one), I wanted to create a test for it and add some extra clean up for the 3 tests that were still creating their own HttpClient (and MinioClient) instances. Those 3 tests depended on `DownloadObjectAsync` and `UploadObjectAsync` methods. 

In the meantime, I discussed with Bala and like the other SDKs, we've decided a functional test for socket leakage is not necessary.

I also got the reviews from you for the unnecessary HttpHandler creation which was turning off TLS and the review about the proxy setting that causes not honoring local system proxies.

So, this PR will make the last 3 test cases to reuse the existing HttpClient instance and fix the wrong TLS and Proxy settings.
I'll change the title to reflect the correct definition of these changes.

harshavardhana (approved):

PR #644, opened 2022-05-26
Fixes multi-space issue in header request parameters

### LINKED ISSUE #419: Presigned GET when specifying header with multiple spaces in value fails with SignatureDoesNotMatch
Consider the following new unit test in `OperationsTest.cs`:

```cs
[TestMethod]
public async Task PresignedGetObjectWithHeadersAndExtraSpaces()
{
    var client = new MinioClient(endpoint:"localhost:9001", "my-access-key", "my-secret-key");

    Mock<IRestClient> restClient = MockRestClient(client.restClient.BaseUrl);
    client.restClient = restClient.Object;

    Dictionary<string, string> reqParams = new Dictionary<string, string>
    {
        {"Response-Content-Disposition", "attachment; filename=\"filename  with spaces.jpg\""},
    };
    var signedUrl = await client.PresignedGetObjectAsync("bucket", "object-name", 3600, reqParams, _requestDate);

    Assert.AreEqual(
        "http://localhost:9001/bucket/object-name?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=my-access-key%2F20200501%2Fus-east-1%2Fs3%2Faws4_request&X-Amz-Date=20200501T154533Z&X-Amz-Expires=3600&X-Amz-SignedHeaders=host&response-content-disposition=attachment%3B%20filename%3D%22filename%20%20with%20spaces.jpg%22&X-Amz-Signature=dc7df0929672f161fdee63fc29a7d827f326cfd6ec994ec68605f13b90273aca",
        signedUrl);
}
```

The `Response-Content-Disposition` header intentionally has multiple sequential spaces.

When encoding, the `Signature` parameter is the equal as there is only one space in the header value, and minio accepts the request.  However, in this case, with two sequential spaces, minio rejects the request.

Maybe minio is not normalizing header values when they are included in the query parameters with the [`signV4TrimAll` function](https://github.com/vadmeste/minio/blob/master/cmd/signature-v4-utils.go#L212)?

I haven't tested this on Amazon's S3 servers, can anyone confirm what the S3 behavior is and what this client should do?

ebozduman:
Fixes #419 
After replacing RestSharp with HttpClient, the issue explained in #419 changed its behavior a little, but the cause of the issue, multiple spaces in request parameters, and the reproduction steps were still causing a problem. The error message now is
```
"System.ArgumentNullException: Value cannot be null. (Parameter 'Unable to download via presigned URL')"
```

### DIFF
--- a/Minio.Functional.Tests/Program.cs
+++ b/Minio.Functional.Tests/Program.cs
@@ -175,7 +175,6 @@ internal class Program
         FunctionalTest.PresignedGetObject_Test3(minioClient).Wait();
         FunctionalTest.PresignedPutObject_Test1(minioClient).Wait();
         FunctionalTest.PresignedPutObject_Test2(minioClient).Wait();
-        FunctionalTest.PresignedGetObject_Test1(minioClient).Wait();
         // FunctionalTest.PresignedPostPolicy_Test1(minioClient).Wait();
 
         // Test incomplete uploads
--- a/Minio/DataModel/BucketArgs.cs
+++ b/Minio/DataModel/BucketArgs.cs
@@ -44,7 +44,7 @@ public abstract class BucketArgs<T> : Args
         {
             if (Headers.ContainsKey(key))
                 Headers.Remove(key);
-            Headers[key] = headers[key];
+            Headers[key] = headers[key].Replace('"', '\"');
         }
 
         return (T)this;
--- a/Minio.Functional.Tests/FunctionalTest.cs
+++ b/Minio.Functional.Tests/FunctionalTest.cs
@@ -4904,7 +4904,7 @@ public class FunctionalTest
             { "expiresInt", expiresInt.ToString() },
             {
                 "reqParams",
-                "response-content-type:application/json,response-content-disposition:attachment;filename=MyDocument.json;"
+                "response-content-type:application/json,response-content-disposition:attachment;filename=  MyDoc u m  e   nt.json ;"
             },
             { "reqDate", reqDate.ToString() }
         };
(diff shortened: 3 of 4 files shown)


Minio.Functional.Tests/Program.cs
```diff
@@ -175,7 +175,6 @@ public static void Main(string[] args)
         FunctionalTest.PresignedGetObject_Test3(minioClient).Wait();
         FunctionalTest.PresignedPutObject_Test1(minioClient).Wait();
         FunctionalTest.PresignedPutObject_Test2(minioClient).Wait();
-        FunctionalTest.PresignedGetObject_Test1(minioClient).Wait();
```
harshavardhana, line 178: why is this test removed?

harshavardhana (changes requested):

Minio.Functional.Tests/Program.cs
ebozduman: A duplicate.
It is also called at the top.

Minio/DataModel/BucketArgs.cs
```diff
@@ -44,7 +44,7 @@ public T WithHeaders(Dictionary<string, string> headers)
         {
             if (Headers.ContainsKey(key))
                 Headers.Remove(key);
-            Headers[key] = headers[key];
+            Headers[key] = headers[key].Replace('"', '\"');
```
ebozduman, line 47: There is no need for this change.
I'll remove it.

ebozduman: Done

harshavardhana (approved):

PR #655, opened 2022-06-24
Remove objectSize validation

haiduong87:
empty file is okay

### DIFF
--- a/Minio/DataModel/ObjectOperationsArgs.cs
+++ b/Minio/DataModel/ObjectOperationsArgs.cs
@@ -1777,9 +1777,6 @@ public class PutObjectArgs : ObjectWriteArgs<PutObjectArgs>
             throw new ArgumentException("Only one of " + nameof(FileName) + " or " + nameof(ObjectStreamData) +
                                         " should be set.");
         if (!string.IsNullOrWhiteSpace(FileName)) utils.ValidateFile(FileName);
-        // Check object size when using stream data
-        if (ObjectStreamData != null && ObjectSize < 0)
-            throw new ArgumentException($"{nameof(ObjectSize)} cannot be less than 0.");
         Populate();
     }
 
--- a/Minio/DataModel/ObjectOperationsArgs.cs
+++ b/Minio/DataModel/ObjectOperationsArgs.cs
@@ -1777,6 +1777,9 @@ public class PutObjectArgs : ObjectWriteArgs<PutObjectArgs>
             throw new ArgumentException("Only one of " + nameof(FileName) + " or " + nameof(ObjectStreamData) +
                                         " should be set.");
         if (!string.IsNullOrWhiteSpace(FileName)) utils.ValidateFile(FileName);
+        // Check object size when using stream data
+        if (ObjectStreamData != null && ObjectSize < 0)
+            throw new ArgumentException($"{nameof(ObjectSize)} cannot be less than 0.");
         Populate();
     }
 
--- a/Minio/DataModel/ObjectOperationsArgs.cs
+++ b/Minio/DataModel/ObjectOperationsArgs.cs
@@ -1777,9 +1777,6 @@ public class PutObjectArgs : ObjectWriteArgs<PutObjectArgs>
             throw new ArgumentException("Only one of " + nameof(FileName) + " or " + nameof(ObjectStreamData) +
                                         " should be set.");
         if (!string.IsNullOrWhiteSpace(FileName)) utils.ValidateFile(FileName);
-        // Check object size when using stream data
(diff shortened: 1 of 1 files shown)


Minio/DataModel/ObjectOperationsArgs.cs
```diff
@@ -1777,9 +1777,6 @@ internal override void Validate()
             throw new ArgumentException("Only one of " + nameof(FileName) + " or " + nameof(ObjectStreamData) +
                                         " should be set.");
         if (!string.IsNullOrWhiteSpace(FileName)) utils.ValidateFile(FileName);
-        // Check object size when using stream data
```
harshavardhana, line 1780: what we expect is we cannot have - ObjectSize < 0

harshavardhana (changes requested):
keep the check add `< 0`

ebozduman (approved):

ebozduman:
Oops! Sorry for the unintended approval. 
@haiduong87 
You need to fix the build failures...

Minio/DataModel/ObjectOperationsArgs.cs
```diff
@@ -1778,8 +1778,8 @@ internal override void Validate()
                                         " should be set.");
         if (!string.IsNullOrWhiteSpace(FileName)) utils.ValidateFile(FileName);
         // Check object size when using stream data
-        if (ObjectStreamData != null && ObjectSize == 0)
-            throw new ArgumentException($"{nameof(ObjectSize)} must be set");
+        if (ObjectStreamData != null && ObjectSize < 0)
+            throw new ArgumentException($"{nameof(ObjectSize)} cannot be less than 0.");
```
ebozduman, line 1782: We set `ObjectSize` to `-1` when `ObjectStreamData` is set. That' why the original error message was saying `{nameof(ObjectSize)} must be set` as `0` size means `ObjectSize` variable is not set.

ebozduman (changes requested):

Minio/DataModel/ObjectOperationsArgs.cs
ebozduman: Just checked one more time. In the current implementation size=0, means size is not defined. But you can `PUT` a file of size `0` by setting its size to `-1`, which seems to be used for unknown size.
I think we don't need this check at all.

ebozduman: @haiduong87 

You also need to make sure to run `regitlint` after making your changes as a last step just before `git push`.

haiduong87: @ebozduman @harshavardhana 
So we don't need to check the size. Just like the 1st commit.

I will make another submit

ebozduman (approved):
LGTM

ebozduman (changes requested):
After digging out about the `-1` unknown size paradigm, it looks like that is our approach also in all other SDKs. So, going back to the starting point and @harshavardhana's comment, you need to put the check back again as there is no change is needed.

In other words, this is our design and there is nothing wrong here.

If you would like to upload zero/0 byte files, then you need to set the size to `-1`.

You can close this PR.

haiduong87:
Hi @ebozduman 

Can you check again?
I modified the test to upload the zero/0 file (with file name)
And it pass the test

image

There're 2 cases for this test:
- I upload file using byte[0] 
- I make empty temp file, and upload.

This lead to a weird case
- I open a filestream, I upload using stream and object size => fail
- I upload using that filename => ok

I think this will confuse users.

ebozduman:
> There're 2 cases for this test:
> 
> * I upload file using byte[0]
> * I make empty temp file, and upload.
> 
> This lead to a weird case
> 
> * I open a filestream, I upload using stream and object size => fail
> * I upload using that filename => ok
> 
> I think this will confuse users.

@haiduong87 ,

MinIO .NET api `PutObjectAsync`, just like in all other supported MinIO SDKs, works as expected and the current behavior is by design. 
So, there is nothing wrong and there is no need for this PR.

Here is how it works:
You can provide the file to be uploaded in 2 ways: either by providing a file name as your source or a stream file.

- **Providing upload data as a filename:**
MinIO will calculate the size of the user provided source file and the upload will proceed without size information. The same for 0-byte files.
Note: Providing the `size` information is useless as it will be ignored.

- **Providing upload data as a stream file**
User is expected to provide the mandatory parameter, the size of the stream. If `size=0`, it means user has not provided the size, hence this is interpreted as an error condition. If the size is unknown, then the size has to be set to `-1`. 
Because of the special meaning of `size=0`, user should set `size=-1` when 0 length stream file is intended to be uploaded.

Please file a documentation issue, if you think this information is not clearly explained in the docs.

Issue #649, opened 2022-06-20
ListenBucketNotificationsAsync usage


DominikHerold:
Hi,

I'm running locally a MinIO instance and implemented the `ListenBucketNotifications` like it is described [here](https://github.com/minio/minio-dotnet/blob/master/Minio.Examples/Cases/ListenBucketNotifications.cs)
I'm uploading a file via the MinIO console but the `onNext` Action does not kick in due to `onCompleted` is already executed. 
Actually `ListenBucketNotificationsAsync` is intended to get informed in case `EventType` is `EventType.ObjectCreatedAll`, isn't it?
Did I missed something or is there another way to get informed when a file is uploaded?
Thanks in advance.

ebozduman:
@DominikHerold 

I cannot reproduce the issue here with my local setup.
Could you please attach here the code you've seen the problem with?

ebozduman:
I reproduced the issue.
It turns out that this is just a simple timing issue as anticipated. So, subscriber needs some time to create notification for the event to be captured.
That is, if you add some sleep time after the event happens, then it will be captured and reported as expected.

Example:
```
using System.Threading;
                :
                :
                await _minio.PutObjectAsync(putObjectArgs);
                Console.WriteLine("Successfully uploaded");
                Thread.Sleep(350);
```

Please give it a try and let us know. You may close the issue if it works.

ebozduman:
@DominikHerold 

Did you have a chance to check it out?

ebozduman:
Closing...
Everything works as expected.

codemnk:
Having same issue. onCompleted is executed almost immediately after observable.Subscribe.
I am pretty sure the issue is in using of async void lambda in
https://github.com/minio/minio-dotnet/blob/f30de0606072b06d2553f8ec24efde4def631ad7/Minio/DataModel/BucketOperationsArgs.cs#L398

ResponseWriter is Action<Stream>. 
Basically response is disposed before stream is processed since calling code does not "await" on async void lambda.
https://devblogs.microsoft.com/pfxteam/potential-pitfalls-to-avoid-when-passing-around-async-lambdas/
https://www.jetbrains.com/help/resharper/AsyncVoidLambda.html

ebozduman:
Thank you @codemnk 
I'll check it out and fix the issue accordingly.

ebozduman:
@codemnk ,

[...]

PR #666, opened 2022-07-20
Listen bucket notification fix

ebozduman:
Fix for issue #649

### DIFF
--- a/Minio/HttpRequestMessageBuilder.cs
+++ b/Minio/HttpRequestMessageBuilder.cs
@@ -60,10 +60,6 @@ internal class HttpRequestMessageBuilder
     public Func<Stream, Task> FunctionResponseWriter { get; set; }
     public HttpMethod Method { get; }
 
-    public void ProcessFunctionResponseWriter(Func<Stream, Task> func, Stream stream)
-    {
-        func(stream).Wait();
-    }
     public HttpRequestMessage Request
     {
         get
@@ -147,6 +143,11 @@ internal class HttpRequestMessageBuilder
     public byte[] Content { get; private set; }
 
     public string ContentTypeKey => "Content-Type";
+
+    public void ProcessFunctionResponseWriter(Func<Stream, Task> func, Stream stream)
+    {
+        func(stream).Wait();
+    }
 
     public void AddHeaderParameter(string key, string value)
     {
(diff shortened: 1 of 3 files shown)


harshavardhana (approved):

Minio/MinioClient.cs
```diff
@@ -529,6 +529,9 @@ public MinioClient WithCredentialsProvider(ClientProvider provider)
             responseResult = new ResponseResult(request, response);
             if (requestMessageBuilder.ResponseWriter != null)
                 requestMessageBuilder.ResponseWriter(responseResult.ContentStream);
+            if (requestMessageBuilder.FunctionResponseWriter != null)
+                requestMessageBuilder.ProcessFunctionResponseWriter(requestMessageBuilder.FunctionResponseWriter,
```
codemnk, line 533: why not simply
`await requestMessageBuilder.FunctionResponseWriter(responseResult.ContentStream)`
?

Minio/HttpRequestMessageBuilder.cs
```diff
@@ -56,6 +57,7 @@ public HttpRequestMessageBuilder(HttpMethod method, Uri requestUri)
 
     public Uri RequestUri { get; set; }
     public Action<Stream> ResponseWriter { get; set; }
+    public Func<Stream, Task> FunctionResponseWriter { get; set; }
```
codemnk, line 60: To make it even more async/await friendly - `Func<Stream, **CancellationToken**, Task>` and `await FunctionResponseWriter(stream, cancellationToken);` on the caller side

Minio/MinioClient.cs
ebozduman: @codemnk ,

Belated thanks for your review comments.
I'll create another PR for your suggested changes when we determine them to be good.

About your comment:
Response writer is not always populated, so we have to check it. Hence we cannot simply make a call to FunctionResponseWriter.

Minio/HttpRequestMessageBuilder.cs
ebozduman: @codemnk ,

Thank you. 
Yes, we can do this.
Either I'll create a sole PR for it or piggyback and include it in the upcoming PR to add 2 new bucket notification tests into the functional tests.

Minio/MinioClient.cs
codemnk: @ebozduman,

I was not suggesting removing null check, I was talking about next line:
instead of using "task.Wait()" (sync call) you better use "await task" (async call)

ebozduman: ok.
Thanks.
The new PR for your review comments is #669.

PR #677, opened 2022-08-15
Fixing issue 673

anthonyMc11:
When the listObjects is called the region was not being calculated ( because the objectName is null) .
This caused a error on s3 to use the perminently redirected address instead of getting the files, now the region is calculated correctly, the code works as expected


ebozduman (approved):
LGTM

ebozduman:
@anthonyMc11 
Please fix the build issues.

About build issue on Linux: 
Please run `dotnet regitlint` and push the lint changes into your PR.

About build failure on Windows:
```
MinIO API responded with message=Storage backend has reached its minimum free drive threshold. Please delete a few objects to proceed.","error":"Minio.Exceptions.UnexpectedMinioException: MinIO API responded with message=Storage backend has reached its minimum free drive threshold. Please delete a few objects to proceed.
```

This could be a false alert/glitch. The docker image used in this test is Windows base OS container, hence we should not have a storage space problem there.
Let's see how the build tests run after pushing the regitlint change.

anthonyMc11:
@ebozduman  I have completed the lint command and updated my PR with the result of it

anthonyMc11:
merged in master

Issue #675, opened 2022-08-12
Inject Ioc-managed HttpClient for MinioClient via constructor


Ruthlessist:
In version 4.0.4 we had the `MinioClient(HttpClient httpClient)` constructor which unfortunately is now marked obsolete:
* Having this function can easily use `services.AddHttpClient<MinioClient>()` to inject HttpClient.
* The parameterless constructor creates an HttpClient instance. If another httpclient instance is subsequently passed in using the builder mode, we will have a garbage. Maybe we can simply set a switch, e.g. `MinioClient(bool createHttpClient = true)`, to control whether to create in the constructor or not.

ebozduman:
PR #681 only addresses the 2nd bullet in the description of this issue above by cleaning up the previously created HttpClient instance, before creating a new one which happens when `WithHttpClient` or `WithProxy` methods are called.

ebozduman:
The team's decision is to go with builder pattern in methods, and this decision is valid for all MinIO implementations where ever it is applicable.
So, the bottom-line is we cannot revive back the methods, which went into obsolete/deprecated mode for this  reason.

Issue #658, opened 2022-06-29
WithProxy is ignored


chrisbecke:
## Issue

We upgraded our own project to the net6.0 framework, and that included upgrading the minio-dotnet SDK to version 4.0.4 As a consequence of this update the Minio SDK now seems to be ignoring .WithProxy().

To get minio-dotnet to route traffic via a proxy it now seems necessary to use WithHttpClient.

## To Reproduce

- Create a new .net6 console application using Visual Studio.
- Enable Docker Compose Support, and add a minio entry as per the docker-compose below.
- Create a reverse proxy using express.js, and add that to the compose.
- Run the project and see that `.WithProxy` is never used, but `.WithHttpClient()` does route through the proxy.

`docker-compose.yml`
```yaml
version: '3.4'

services:
  copytest:
    image: ${DOCKER_REGISTRY-}copytest
    build:
      context: .
      dockerfile: CopyTest/Dockerfile
    depends_on:
    - proxy

  proxy:
    image: ${DOCKER_REGISTRY-}proxy
    build:
      context: proxy
      target: dev
    volumes:
    - ./proxy:/src
    - /src/node_modules
    depends_on:
    - minio

  minio:
    image: minio/minio:latest
    command: server /data --console-address :9001
    volumes:
    - /data
    ports:
    - 8080:9001

  mc:
    image: minio/mc:latest
    profiles: [ "tools" ]
    working_dir: /work
    volumes:
    - ./.mc:/root/.mc
    - ./:/work
```

`Program.cs`
```csharp
using Minio;
using Minio.DataModel;
using System.Net;

var minio = new MinioClient()
                        .WithCredentials("minioadmin","minioadmin")
                        .WithEndpoint("minio:9000")
//                        .WithHttpClient( new HttpClient(new HttpClientHandler(){ Proxy = new WebProxy("proxy", 3000) }))
                        .WithProxy(new WebProxy("proxy", 3000))
                        .Build();

string fromBucketName = "uploads";
string fromObjectName = "test.txt";
string destBucketName = "staging";
string destObjectName = "test.txt";
ServerSideEncryption sseSrc = null;
ServerSideEncryption sseDest = null;

var cpSrcArgs = new CopySourceObjectArgs()
    .WithBucket(fromBucketName)
    .WithObject(fromObjectName)
    .WithServerSideEncryption(sseSrc);
var cpArgs = new CopyObjectArgs()
    .WithBucket(destBucketName)
    .WithObject(destObjectName)
[...]

PR #681, opened 2022-08-23
Fixes WithProxy(#658) and leftover HttpClient(#675) issues

ebozduman:
Fixes 2 issues:
#658 : WithProxy is ignored
#675 : Inject Ioc-managed HttpClient for MinioClient via constructor


Ruthlessist (commented):
I don't think this is a good solution, refer to PR #683.

Issue #685, opened 2022-09-02
ObjectNotfoundException in RemoveObjectAsync tooltip needs to be removed


dthemg:
When removing an object using `RemoveObejctAsync` no error is thrown
```
await _minioClient.RemoveObjectAsync(new RemoveObjectArgs().WithBucket("my-bucket").WithObject(Guid.NewGuid().ToString()));
```
It is however stated in the tooltip that an exception should be thrown if the object isn't found:

![image](https://user-images.githubusercontent.com/13008236/188171804-b0da9467-451d-4908-9429-be2f00e9c20c.png)

ebozduman:
@dthemg 

This is a tooltip bug. `"ObjectNotFound"` exception does not belong there and needs to be removed.

So, the existing behavior is correct.

ebozduman:
Fixed by #690

PR #690, opened 2022-09-19
Removes the tooltip "ObjectNotfoundException" that will never happen

ebozduman:
Fix for #685

### DIFF
--- a/Minio.Functional.Tests/Program.cs
+++ b/Minio.Functional.Tests/Program.cs
@@ -91,8 +91,8 @@ internal class Program
         // "Listening for bucket notification is specific only to `minio`
         // server endpoints".
         FunctionalTest.ListenBucketNotificationsAsync_Test1(minioClient).Wait();
-        // FunctionalTest.ListenBucketNotificationsAsync_Test2(minioClient).Wait();
-        // FunctionalTest.ListenBucketNotificationsAsync_Test3(minioClient).Wait();
+        FunctionalTest.ListenBucketNotificationsAsync_Test2(minioClient).Wait();
+        FunctionalTest.ListenBucketNotificationsAsync_Test3(minioClient).Wait();
 
         // Check if bucket exists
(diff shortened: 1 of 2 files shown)


Minio.Functional.Tests/Program.cs
```diff
@@ -91,8 +91,8 @@ public static void Main(string[] args)
         // "Listening for bucket notification is specific only to `minio`
         // server endpoints".
         FunctionalTest.ListenBucketNotificationsAsync_Test1(minioClient).Wait();
-        FunctionalTest.ListenBucketNotificationsAsync_Test2(minioClient).Wait();
-        FunctionalTest.ListenBucketNotificationsAsync_Test3(minioClient).Wait();
+        // FunctionalTest.ListenBucketNotificationsAsync_Test2(minioClient).Wait();
+        // FunctionalTest.ListenBucketNotificationsAsync_Test3(minioClient).Wait();
```
harshavardhana, line 95: why is this being commented out? @ebozduman

harshavardhana (changes requested):

Minio.Functional.Tests/Program.cs
ebozduman: There is an intermittent issue, #691, in these 2 functional tests due to the way `Thread.Sleep()` is used in them to wait for notifications to happen. 

I hit this issue during build tests (git workflow runs) when I was working on PR #687 (s3-zip support). Since these 2 tests are not related to the this PR or #687, I did not want to withhold them because of an unrelated problem. 

Hence the temporary comments on these 2 tests until #691 is fixed.

I am currently working on the #691 and the fix for it will come with these comments removed for these 2 tests.

ebozduman:
@harshavardhana 

With the merge of PR #693, I removed the comments from the 2 bucket notification tests.

harshavardhana (approved):

PR #692, opened 2022-09-20
Updating documentation links to new paths

djwfyi:
Correcting README links to the documentation to point to updated paths for the updated docs site.

PR #693, opened 2022-09-24
Fixes random failures in bucket notification tests

### LINKED ISSUE #691: Sometimes event notifications are missed during git workflow func test runs.
2 notification tests run successfully when functional tests are run by `dotnet run` command locally, but fail to run sporadically during git build workflows.
It turned out to be the `Thread.Sleep()` command used to wait for the event to happen is wrongly used and sometimes (~20% of the runs) it caused the event to be missed. This issue needs to be addressed to make the tests run to completion successfully 100% of the times.

Note: Using `Task.Delay()` helps, but tests still fail 5% of the times since the event could not be observed.

ebozduman:
Fixes #691

Sometimes the event is created so fast even before the subscriber is ready to capture the event, hence random failures. So, adding delay and sleep in the right places in the tests resolved the issue for almost always, although theoretically there is always a very low but still a chance to see this problem.

### DIFF
--- a/Minio.Functional.Tests/FunctionalTest.cs
+++ b/Minio.Functional.Tests/FunctionalTest.cs
@@ -2543,7 +2543,8 @@ public class FunctionalTest
         var startTime = DateTime.Now;
         var events = new List<EventType>();
         events.Add(EventType.ObjectCreatedAll);
-        var rxEvents = new List<NotificationEvent>();
+        var rxEventData = new MinioNotificationRaw("");
+        var rxEventsList = new List<NotificationEvent>();
         IDisposable subscription = null;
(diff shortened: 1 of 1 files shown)


ebozduman:
Build functional test still fails on windows environment.
It is kind of surprising still seeing the failure as I ran the same set of bucket notification tests (3 of them) back to back 100 times on a VM Windows 2019 server and I did not see a single failure.
I am still looking into it.

ebozduman:
@harshavardhana 

This fix is needed for PR #692 and PR #694. Their build tests fail with the same reason.

harshavardhana (approved):

Issue #656, opened 2022-06-28
Error during setting .WithLifecycleConfiguration


peterczabanski:
Hi, settings like:
` SetBucketLifecycleArgs args = new SetBucketLifecycleArgs()
                    .WithBucket(bucketName)
                    .WithLifecycleConfiguration(
                        new LifecycleConfiguration(new List<LifecycleRule>(new[] { new LifecycleRule
                            {
                                AbortIncompleteMultipartUploadObject = null,
                                ID = "Delete imagines after 31 days",
                                Expiration = null,
                                TransitionObject = null,
                                Filter = null,
                                NoncurrentVersionExpirationObject = new NoncurrentVersionExpiration(31),
                                NoncurrentVersionTransitionObject = null,
                                Status = LifecycleRule.LIFECYCLE_RULE_STATUS_ENABLED
                            }
                        })));`

return error:

`API: PutBucketLifecycle(bucket=trackingid)
Time: 11:59:46 UTC 06/28/2022
DeploymentID: 3b10a44d-fa41-434a-a056-ffef5d19d9f3
RequestID: 16FCC704C6A47418
RemoteHost: 172.17.0.1
Host: 10.170.44.31:19000
UserAgent: MinIO (Linux 5.10.16.3-microsoft-standard-WSL2 #1 SMP Fri Apr 2 22:23:49 UTC 2021;X64) minio-dotnet/4.0.3.0
Error: The XML you provided was not well-formed or did not validate against our published schema (lifecycle.Error)
       5: internal/logger/logger.go:278:logger.LogIf()
       4: cmd/api-errors.go:2123:cmd.toAPIErrorCode()
       3: cmd/api-errors.go:2148:cmd.toAPIError()
       2: cmd/bucket-lifecycle-handlers.go:78:cmd.objectAPIHandlers.PutBucketLifecycleHandler()
       1: net/http/server.go:2084:http.HandlerFunc.ServeHTTP()
`

yadanilov19:
Hi! A have the same problem.
Code:
```csharp
 var lifeCycle = new SetBucketLifecycleArgs()
                .WithBucket("test")
                .WithLifecycleConfiguration(
                    new LifecycleConfiguration()
                    {
                        Rules = new List<LifecycleRule>()
                        {
                            new()
                            {
                                ID = "AutoRemoveOldFiles",
                                Status = LifecycleRule.LIFECYCLE_RULE_STATUS_ENABLED,
[...]

PR #722, opened 2022-12-05
Fix for issue during setting .WithLifecycleConfiguration (#656)

MarDipp:
Fixes issue [Error during setting .WithLifecycleConfiguration #656](https://github.com/minio/minio-dotnet/issues/656).

The XML node NoncurrentVersionExpiration was empty because internal properties are not serialized.

### DIFF
--- a/Minio/DataModel/ILM/LifecycleRule.cs
+++ b/Minio/DataModel/ILM/LifecycleRule.cs
@@ -33,8 +33,11 @@ public class LifecycleRule
     public static readonly string LIFECYCLE_RULE_STATUS_ENABLED = "Enabled";
     public static readonly string LIFECYCLE_RULE_STATUS_DISABLED = "Disabled";
 
+    private RuleFilter _ruleFilter;
+
     public LifecycleRule()
     {
+        _ruleFilter = new RuleFilter();
     }
 
     public LifecycleRule(AbortIncompleteMultipartUpload abortIncompleteMultipartUpload, string id,
@@ -67,7 +70,22 @@ public class LifecycleRule
     public Transition TransitionObject { get; set; }
 
     [XmlElement("Filter", IsNullable = true)]
-    public RuleFilter Filter { get; set; }
+    public RuleFilter Filter
+    {
+        get { return _ruleFilter; }
+        set
+        {
+            // The filter must not be missing, even if it is empty.
+            if (value == null)
+            {
+                _ruleFilter = new RuleFilter();
+            }
+            else
+            {
+                _ruleFilter = value;
+            }
+        }
+    }
 
     [XmlElement("NoncurrentVersionExpiration", IsNullable = true)]
     public NoncurrentVersionExpiration NoncurrentVersionExpirationObject { get; set; }
@@ -75,5 +93,6 @@ public class LifecycleRule
     [XmlElement("NoncurrentVersionTransition", IsNullable = true)]
     public NoncurrentVersionTransition NoncurrentVersionTransitionObject { get; set; }
 
-    [XmlElement("Status")] public string Status { get; set; }
+    [XmlElement("Status")]
+    public string Status { get; set; }
 }
--- a/Minio/DataModel/ILM/LifecycleRule.cs
+++ b/Minio/DataModel/ILM/LifecycleRule.cs
@@ -72,18 +72,14 @@ public class LifecycleRule
     [XmlElement("Filter", IsNullable = true)]
     public RuleFilter Filter
     {
-        get { return _ruleFilter; }
+        get => _ruleFilter;
         set
         {
             // The filter must not be missing, even if it is empty.
             if (value == null)
-            {
                 _ruleFilter = new RuleFilter();
-            }
             else
-            {
                 _ruleFilter = value;
-            }
         }
     }
 
@@ -93,6 +89,5 @@ public class LifecycleRule
     [XmlElement("NoncurrentVersionTransition", IsNullable = true)]
     public NoncurrentVersionTransition NoncurrentVersionTransitionObject { get; set; }
 
-    [XmlElement("Status")]
-    public string Status { get; set; }
+    [XmlElement("Status")] public string Status { get; set; }
 }
--- a/Minio/DataModel/ILM/NoncurrentVersionExpiration.cs
+++ b/Minio/DataModel/ILM/NoncurrentVersionExpiration.cs
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+using System;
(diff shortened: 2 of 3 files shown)


harshavardhana (approved):

ebozduman:
Reviewing...

ebozduman (changes requested):
I ran with @peterczabanski 's configuration given in the issue tobe fixed, #656, and `SetBucketLifecycleAsync` api still fails the same way with the same error message.
Your fix looks good, but it seems like the fix needs more changes.

One thing I noticed is; minio-go has 2 properties for `NoncurrentVersionExpiration`
```
<NoncurrentVersionExpiration>     
            <NewerNoncurrentVersions>5</NewerNoncurrentVersions>
            <NoncurrentDays>365</NoncurrentDays>    
</NoncurrentVersionExpiration> 
```
i.e `NewerNoncurrentVersions` is missing.

MarDipp:
> I ran with @peterczabanski 's configuration given in the issue tobe fixed, #656, and `SetBucketLifecycleAsync` api still fails the same way with the same error message. Your fix looks good, but it seems like the fix needs more changes.
> 
> One thing I noticed is; minio-go has 2 properties for `NoncurrentVersionExpiration`
> 
> ```
> <NoncurrentVersionExpiration>     
>             <NewerNoncurrentVersions>5</NewerNoncurrentVersions>
>             <NoncurrentDays>365</NoncurrentDays>    
> </NoncurrentVersionExpiration> 
> ```
> 
> i.e `NewerNoncurrentVersions` is missing.

The remaining problem with @peterczabanski 's configuration is the part `Filter = null`. [The Filter must not be empty](https://docs.aws.amazon.com/AmazonS3/latest/userguide/intro-lifecycle-rules.html#intro-lifecycle-rules-filter). To apply the rule to all files an empty prefix can be used: `Filter = new RuleFilter() { Prefix = String.Empty }`

This configuration will work:
`SetBucketLifecycleArgs args = new SetBucketLifecycleArgs() .WithBucket(bucketName) .WithLifecycleConfiguration( new LifecycleConfiguration(new List<LifecycleRule>(new[] { new LifecycleRule { AbortIncompleteMultipartUploadObject = null, ID = "Delete imagines after 31 days", Expiration = null, TransitionObject = null, Filter = new RuleFilter() { Prefix = String.Empty }, NoncurrentVersionExpirationObject = new NoncurrentVersionExpiration(31), NoncurrentVersionTransitionObject = null, Status = LifecycleRule.LIFECYCLE_RULE_STATUS_ENABLED } })));`

I'm not sure how this can be fixed. Perhaps an error message can be helpful.

ebozduman:
@MarDipp ,

> This configuration will work:
SetBucketLifecycleArgs args = new SetBucketLifecycleArgs() .WithBucket(bucketName) .WithLifecycleConfiguration( new LifecycleConfiguration(new List<LifecycleRule>(new[] { new LifecycleRule { AbortIncompleteMultipartUploadObject = null, ID = "Delete imagines after 31 days", Expiration = null, TransitionObject = null, Filter = new RuleFilter() { Prefix = String.Empty }, NoncurrentVersionExpirationObject = new NoncurrentVersionExpiration(31), NoncurrentVersionTransitionObject = null, Status = LifecycleRule.LIFECYCLE_RULE_STATUS_ENABLED } })));

Yes. This configuration works.
But we have to fix the `null` value issue for the `Filter` tag. Otherwise, it needs to be documented and it sounds pretty bad to me.

We also need to add the missing `NewerNoncurrentVersions` tag in `NoncurrentVersionExpiration`.

MarDipp:
> ...
> Yes. This configuration works. But we have to fix the `null` value issue for the `Filter` tag. Otherwise, it needs to be documented and it sounds pretty bad to me.
> 
> We also need to add the missing `NewerNoncurrentVersions` tag in `NoncurrentVersionExpiration`.

I added the property `NewerNoncurrentVersions` like in minio-go.
Additionally it is ensured that the filter is never null. Now @peterczabanski 's configuration works without change.

ebozduman (changes requested):
I think there is one last issue:

`NoncurrentVersionExpirationObject` has become a mandatory tag and it has to be defined.
Try it with a `null` value and with no `NoncurrentVersionExpirationObject` definition in your Rule.

MarDipp:
> I think there is one last issue:
> 
> `NoncurrentVersionExpirationObject` has become a mandatory tag and it has to be defined. Try it with a `null` value and with no `NoncurrentVersionExpirationObject` definition in your Rule.

Have you added an other condition?

I don´t think `NoncurrentVersionExpirationObject` is mandatory. But you need either `Expiration`, `TransitionObject`, `NoncurrentVersionExpirationObject` or `NoncurrentVersionTransitionObject` filled. If all of these objects are missing, an XML error is thrown. [Here is the code of the Rule Validation](https://github.com/minio/minio/blob/6511021fbe65148145390894a34143fee3c053f6/internal/bucket/lifecycle/rule.go#L161)

This configuration has no `NoncurrentVersionExpirationObject`, but works anyway: 
`SetBucketLifecycleArgs setPolicyArgs = new SetBucketLifecycleArgs().WithBucket(bucketName).WithLifecycleConfiguration(new LifecycleConfiguration(new List<LifecycleRule>(new[] { new LifecycleRule { AbortIncompleteMultipartUploadObject = null, ID = "Test", Expiration = new Expiration(DateTime.UtcNow.AddYears(10)), TransitionObject = null, Filter = null, NoncurrentVersionExpirationObject = null, NoncurrentVersionTransitionObject = null, Status = LifecycleRule.LIFECYCLE_RULE_STATUS_ENABLED } })));`

ebozduman:
I see. At this point in the code, there is only one required tag, `"Status"` and the following is one of the minimum configs:
```
LifecycleConfiguration lfc = new LifecycleConfiguration(
    new List<LifecycleRule>(new[]
    { new LifecycleRule
        { 
            NoncurrentVersionExpirationObject = new NoncurrentVersionExpiration(31, 5), 
Glacier Flexible Retrieval"), 
            Status = LifecycleRule.LIFECYCLE_RULE_STATUS_ENABLED
        }
    }
));
```

while the following config fails:
```
LifecycleConfiguration lfc = new LifecycleConfiguration(
    new List<LifecycleRule>(new[]
    { new LifecycleRule
        { 
            Status = LifecycleRule.LIFECYCLE_RULE_STATUS_ENABLED
        }
    }
));

I guess, at least one of the tags, either transitions or expirations, needs to be provided with values.

So, I guess we can stop here.

LGTM

ebozduman (approved):

PR #723, opened 2022-12-07
Adds progress for GetObjectAsync and PutObjectAsync methods

lukashev-di:


ebozduman (changes requested):
@lukashev-di 

Please add a new functional test for this new feature.
You can have one based on one of the `PutObject_Test<N>` tests in `FunctionalTest.cs` file.

Also run `dotnet regitlint` just before you create your commit to be able to include the lint changes in your commit. This will fix the Linux build failure.

ebozduman (changes requested):
We also need to make sure the progress is printed out as `100` when a small object/file is uploaded. I see `0` percent progress after uploading a `1MB` file.

PR #728, opened 2022-12-17
using interface IReadOnlyDictionary instead of a dicrionary in TagSet and Tagging

danilenko-a:
The dictionary is used only for enumeration, so the interface IReadOnlyDictionary is enough


ebozduman (approved):
LGTM

PR #731, opened 2023-01-12
Adds return values for PutObjectAsync api

### LINKED ISSUE #707: No return information after uploading
After uploading, there is no return information ObjectStat. VersionId is available only when downloading. Why does Java have return values

ebozduman:
Fixes #707 
PutObjectAsync returns `Etag`, `Size` and `Name` of the uploaded object.

### DIFF
--- a/Minio/ApiEndpoints/ObjectOperations.cs
+++ b/Minio/ApiEndpoints/ObjectOperations.cs
@@ -821,7 +821,9 @@ public partial class MinioClient : IObjectOperations
     {
         //Skipping validate as we need the case where stream sends 0 bytes
         var requestMessageBuilder = await CreateRequest(args).ConfigureAwait(false);
-        using var response = await ExecuteTaskAsync(NoErrorHandlers, requestMessageBuilder, cancellationToken: cancellationToken).ConfigureAwait(false);
+        using var response =
+            await ExecuteTaskAsync(NoErrorHandlers, requestMessageBuilder, cancellationToken: cancellationToken)
+                .ConfigureAwait(false);
         return new PutObjectResponse(response.StatusCode, response.Content, response.Headers,
             args.ObjectSize, args.ObjectName);
     }
@@ -1058,9 +1060,10 @@ public partial class MinioClient : IObjectOperations
     {
         args?.Validate();
         var requestMessageBuilder = await CreateRequest(args).ConfigureAwait(false);
-        using ResponseResult response = await ExecuteTaskAsync(NoErrorHandlers, requestMessageBuilder, cancellationToken: cancellationToken)
-            .ConfigureAwait(false);
-        return new PutObjectResponse(response.StatusCode, response.Content, response.Headers, response.Content.Length,
+        using var response =
+            await ExecuteTaskAsync(NoErrorHandlers, requestMessageBuilder, cancellationToken: cancellationToken)
+                .ConfigureAwait(false);
+        return new PutObjectResponse(response.StatusCode, response.Content, response.Headers, -1,
             args.ObjectName);
     }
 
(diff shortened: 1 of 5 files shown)


Minio/DataModel/ObjectOperationsResponse.cs
```diff
@@ -215,26 +215,28 @@ internal NewMultipartUploadResponse(HttpStatusCode statusCode, string responseCo
...
 {
-    internal string Etag;
+    public string Etag;
+    public string ObjectName;
+    public string Size;
 
-    internal PutObjectResponse(HttpStatusCode statusCode, string responseContent,
-        Dictionary<string, string> responseHeaders)
+    public PutObjectResponse(HttpStatusCode statusCode, string responseContent,
+        Dictionary<string, string> responseHeaders, string size, string name)
```
pbolduc, line 225: It would be useful to have the response headers exposed for end user interrogation.

Minio/ApiEndpoints/ObjectOperations.cs
```diff
@@ -1154,14 +1158,14 @@ private async Task RemoveUploadAsync(RemoveUploadArgs args, CancellationToken ca
...
+    private async Task<PutObjectResponse> PutObjectSinglePartAsync(PutObjectArgs args,
         CancellationToken cancellationToken = default)
     {
         //Skipping validate as we need the case where stream sends 0 bytes
         var requestMessageBuilder = await CreateRequest(args).ConfigureAwait(false);
         using var response = await ExecuteTaskAsync(NoErrorHandlers, requestMessageBuilder, cancellationToken);
-        var putObjectResponse = new PutObjectResponse(response.StatusCode, response.Content, response.Headers);
-        return putObjectResponse.Etag;
+        return new PutObjectResponse(response.StatusCode, response.Content, response.Headers,
+            args.ObjectSize.ToString(), args.ObjectName);
```
pbolduc, line 1168: Why is `ObjectSize` being converted to string? Just keep it as a `long`.

Minio/ApiEndpoints/ObjectOperations.cs
```diff
@@ -1388,13 +1395,15 @@ private async Task<object> CopyObjectRequestAsync(CopyObjectRequestArgs args, Ca
...
     /// <exception cref="AccessDeniedException">For encrypted copy operation, Access is denied if the key is wrong</exception>
-    private async Task CompleteMultipartUploadAsync(CompleteMultipartUploadArgs args,
+    private async Task<PutObjectResponse> CompleteMultipartUploadAsync(CompleteMultipartUploadArgs args,
         CancellationToken cancellationToken)
     {
         args.Validate();
         var requestMessageBuilder = await CreateRequest(args).ConfigureAwait(false);
         using var response = await ExecuteTaskAsync(NoErrorHandlers, requestMessageBuilder, cancellationToken)
             .ConfigureAwait(false);
+        return new PutObjectResponse(response.StatusCode, response.Content, response.Headers, "unknown",
```
pbolduc, line 1405: You could use `-1` for unknown instead of using a string to represent the length. Do we really need to expose the length on put object? Does the S3 put object / upload APIs return the length? To keep things simpler, would it be too much to ask the user to call a stat or something to get the length if they didn't know how much data they uploaded. Using a string like this is a code smell and is hard to remove once released.

ebozduman: Done

Minio/ApiEndpoints/ObjectOperations.cs
ebozduman: Done

Minio/DataModel/ObjectOperationsResponse.cs
ebozduman: Done

Minio/ApiEndpoints/ObjectOperations.cs
```diff
@@ -572,8 +573,8 @@ public async Task PutObjectAsync(PutObjectArgs args, CancellationToken cancellat
             args = args.WithRequestBody(bytes)
                 .WithStreamData(null)
                 .WithObjectSize(bytesRead);
-            await PutObjectSinglePartAsync(args, cancellationToken).ConfigureAwait(false);
-            return;
+            var putObjResponse = await PutObjectSinglePartAsync(args, cancellationToken).ConfigureAwait(false);
```
martijn00, line 576: Just return the result directly. Don't assign it first

martijn00 (approved):

harshavardhana (approved):

PR #736, opened 2023-02-01
WIP: memory management improvements

ebozduman:
Replacing regular memory allocations with `ArrayPool` when uploading big files.

### DIFF
--- a/Minio/ApiEndpoints/ObjectOperations.cs
+++ b/Minio/ApiEndpoints/ObjectOperations.cs
@@ -16,6 +16,7 @@
  */
 
 using System;
+using System.Buffers;
 using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
@@ -576,15 +577,18 @@ public partial class MinioClient : IObjectOperations
         // Upload object in single part if size falls under restricted part size.
         if (args.ObjectSize < Constants.MinimumPartSize && args.ObjectSize >= 0 && args.ObjectStreamData != null)
         {
-            var bytes = await ReadFullAsync(args.ObjectStreamData, (int)args.ObjectSize).ConfigureAwait(false);
-            var bytesRead = bytes == null ? 0 : bytes.Length;
-            if (bytesRead != (int)args.ObjectSize)
+            var shared = ArrayPool<byte>.Shared;
+            var bytes = shared.Rent((int)args.ObjectSize);
+            bytes = await ReadFullAsync(args.ObjectStreamData, (int)args.ObjectSize).ConfigureAwait(false);
+            var bytesRead = bytes == null ? 0 : bytes.Count();
+            if (bytesRead < (int)args.ObjectSize)
                 throw new UnexpectedShortReadException(
-                    $"Data read {bytesRead} is shorter than the size {args.ObjectSize} of input buffer.");
-            args = args.WithRequestBody(bytes)
+                    $"Data read {bytesRead} is less than the size {args.ObjectSize} of input buffer.");
+            args = args.WithRequestBody(bytes, (int)args.ObjectSize)
                 .WithStreamData(null)
-                .WithObjectSize(bytesRead);
+                .WithObjectSize(args.ObjectSize);
             await PutObjectSinglePartAsync(args, cancellationToken).ConfigureAwait(false);
+            shared.Return(bytes);
             return;
         }
 
@@ -610,7 +614,7 @@ public partial class MinioClient : IObjectOperations
             .WithContentType(args.ContentType)
             .WithUploadId(uploadId)
             .WithStreamData(args.ObjectStreamData)
-            .WithRequestBody(args.RequestBody)
+            .WithRequestBody(args.RequestBody, (int)args.ObjectSize)
             .WithHeaders(args.Headers);
         Dictionary<int, string> etags = null;
         // Upload file contents.
@@ -624,14 +628,11 @@ public partial class MinioClient : IObjectOperations
                     .WithStreamData(fileStream)
                     .WithObjectSize(fileStream.Length)
                     .WithRequestBody(null);
-                etags = await PutObjectPartAsync(putObjectPartArgs, cancellationToken).ConfigureAwait(false);
             }
         }
+
         // Upload stream contents
-        else
-        {
-            etags = await PutObjectPartAsync(putObjectPartArgs, cancellationToken).ConfigureAwait(false);
-        }
+        etags = await PutObjectPartAsync(putObjectPartArgs, cancellationToken).ConfigureAwait(false);
 
         var completeMultipartUploadArgs = new CompleteMultipartUploadArgs()
             .WithBucket(args.BucketName)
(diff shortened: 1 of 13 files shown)


Minio/DataModel/ObjectArgs.cs
```diff
@@ -23,16 +25,23 @@ public abstract class ObjectArgs<T> : BucketArgs<T>
...
     }
 
-    public T WithRequestBody(byte[] data)
+    public T WithRequestBody(byte[] data, int exactBodySize = 0)
     {
-        RequestBody = data;
+        if (data is null) return (T)this;
+
+        ExactBodySize = exactBodySize;
+        RequestBody = new byte[exactBodySize];
```
pbolduc, line 41: Since this is a public API, be sure to check the parameter for valid values. If the value is less than zero, then `throw new ArgumentOutOfRangeException(nameof(exactBodySize), "Exact body size must be greater or equal to zero.")`

Minio/Helper/utils.cs
```diff
@@ -235,12 +235,15 @@ public static bool IsValidExpiry(int expiryInt)
         return expiryInt > 0 && expiryInt <= Constants.DefaultExpiryTime;
     }
 
-    internal static string getMD5SumStr(byte[] key)
+    internal static string getMD5SumStr(byte[] key, int exactBodySize = 0)
     {
+        var keyCopy = new byte[exactBodySize];
```
pbolduc, line 240: Have you considered using an ArrayPool<byte> from `System.Buffers`?

```cs
        var keyCopy = ArrayPool<byte>.Shared.Rent(exactBodySize);

        try
        {
            var hashedBytes = MD5
                .Create()
                .ComputeHash(key);

            return Convert.ToBase64String(hashedBytes);
        }
        finally
        {
            ArrayPool<byte>.Shared.Return(keyCopy);
        }
```

Minio/ApiEndpoints/ObjectOperations.cs
```diff
@@ -1195,18 +1197,21 @@ private async Task RemoveUploadAsync(RemoveUploadArgs args, CancellationToken ca
         var etags = new Dictionary<int, string>();
         for (partNumber = 1; partNumber <= partCount; partNumber++)
         {
-            var dataToCopy = await ReadFullAsync(args.ObjectStreamData, (int)partSize).ConfigureAwait(false);
-            if (dataToCopy == null && numPartsUploaded > 0) break;
+            var shared = ArrayPool<byte>.Shared;
```
pbolduc, line 1200: but the block below in a `try { } finally { ArrayPool<byte>.Shared.Return(shared); }` to avoid memory leak if an exception is thrown.

Minio/ApiEndpoints/ObjectOperations.cs
```diff
@@ -1609,26 +1614,32 @@ private async Task<object> CopyObjectRequestAsync(CopyObjectRequestArgs args, Ca
...
+        var shared = ArrayPool<byte>.Shared;
+        var curData = shared.Rent(currentPartSize - totalRead);
 
-        if (totalRead == 0) return null;
+        try
+        {
+            while (totalRead != currentPartSize &&
+                   totalRead < currentPartSize)
+            {
+                curData = shared.Rent(currentPartSize - totalRead);
```
pbolduc, line 1626: This is a memory leak. You are not freeing the rented buffer from line 1619.

Minio/ApiEndpoints/ObjectOperations.cs
```diff
@@ -1609,26 +1614,32 @@ private async Task<object> CopyObjectRequestAsync(CopyObjectRequestArgs args, Ca
...
+                totalRead += curRead;
+            }
 
-        if (totalRead == currentPartSize) return result;
+            if (totalRead == 0) return null;
 
-        var truncatedResult = new byte[totalRead];
-        for (var i = 0; i < totalRead; i++) truncatedResult[i] = result[i];
-        return truncatedResult;
+            return curData;
```
pbolduc, line 1634: You can't return a rented buffer that is returned the the pool below on line 1638.

Minio/DataModel/ObjectArgs.cs
```diff
@@ -23,16 +25,23 @@ public abstract class ObjectArgs<T> : BucketArgs<T>
...
+    internal int ExactBodySize { get; set; }
 
     public T WithObject(string obj)
     {
         ObjectName = obj;
         return (T)this;
     }
 
-    public T WithRequestBody(byte[] data)
+    public T WithRequestBody(byte[] data, int exactBodySize = 0)
```
pbolduc, line 36: Instead of using `exactBodySize` as a parameter, it would be more consistent to use `count` similar to functions like Buffer.BlockCopy.

pbolduc:
I wonder if you have considered using [Microsoft.IO.RecyclableMemoryStream](https://github.com/microsoft/Microsoft.IO.RecyclableMemoryStream). I would suggest embedding the source to this library as an internal implementation. 
> `RecyclableMemoryStream` improves GC performance by ensuring that the larger buffers used for the streams are put into the gen 2 heap and stay there forever. This should cause full collections to happen less frequently. If you pick buffer sizes above 85,000 bytes, then you will ensure these are placed on the large object heap, which is touched even less frequently by the garbage collector.

Minio/DataModel/ObjectArgs.cs
ebozduman: @pbolduc 

Ok. Will do. 

...and thank you for all your review commands.
btw, there are more changes coming up I introduced while I was testing the memory allocation during uploads for different sizes of objects.

Minio/ApiEndpoints/ObjectOperations.cs
ebozduman: Yes, agreed.
That is one of the incoming changes.

Minio/ApiEndpoints/ObjectOperations.cs
ebozduman: Yes. This part of the code is not finalized yet and might change completely without any `while` if possible.
We just need to make sure there is no regression.

Minio/ApiEndpoints/ObjectOperations.cs
ebozduman: Yes.
This is where I have a `try{} / finally{}` structure and a `shared.Return` in the coming commit.

Minio/Helper/utils.cs
ebozduman: The problem with `ArrayPool` is that it does not give you the exact size you've requested, e.g. if you request an array of size `13`, the pool may give you probably a 16 bytes array or a bigger size depending on pool's availability. Hence, the `exactBodySize` variable is set to keep track of the correct size. If the rented array comes in the exact desired size, then the `exactBodySize` vaiable is set to `0`.
The bottom line is; returning an array by `ArrayPool` in exact/requested size is not guaranteed. It is a drawback in some cases, but it is one of the reasons `ArrayPool` has a smaller/better memory foot print. 

In our case, we don't mind carrying the content in a bigger size array, but only until the time when we calculate the authentication signature for the http request. So, at the latest point just before the http request's authentication signature is calculated, we had to create the correct sized array somehow; either in the memory as a byte[] array or on the disk whenever it is possible. 

The code you've commented on here is where we calculate the `md5sum` of the http content. So, this is where we need the exact sized byte array.

martijn00:
@ebozduman you should either totally rework this PR since the extensive use of Span now, or abandon it. The amount of allocations and copy of memory should already be a lot less now.

ebozduman:
@martijn00 ,
I agree.
We need to re-evaluate our memory foot print after the extensive changes.
Closing at this point.

PR #739, opened 2023-02-03
Response header processing

pbolduc:
This PR tidies up the response header processing. HTTP message headers are [case-insensitive](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2) as per the HTTP specification. In .NET, we can create a dictionary with a case-insensitive key comparer like `new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase)`. When looking up values in this dictionary, the key lookups are case insensitive. This helps when trying to find a key like `ETag` case-insensitively. The lookups can be simplified.

In addition, this changes the type of `ResponseResult.Headers` from `Dictionary<string, string>` to `IReadOnlyDictionary<string, string>`. Read only dictionary only has the members for getting values from the key, the add/update members are not included. Note that `Dictionary<Tkey, Tvalue>` implements `IReadOnlyDictionary<Tkey, Tvalue>`. By using `IReadOnlyDictionary<string, string>` for the  `ResponseResult.Headers`, there is no easy way for callers to manipulate the response headers by accident.

### DIFF
--- a/Minio/MinioClient.cs
+++ b/Minio/MinioClient.cs
@@ -646,17 +646,14 @@ public partial class MinioClient : IMinioClient
         MinioException error = null;
         var errorResponse = new ErrorResponse();
 
-        foreach (var parameter in response.Headers)
-        {
-            if (parameter.Key.Equals("x-amz-id-2", StringComparison.CurrentCultureIgnoreCase))
-                errorResponse.HostId = parameter.Value;
+        if (response.Headers.TryGetValue("x-amz-id-2", out string hostId))
+            errorResponse.HostId = hostId;
 
-            if (parameter.Key.Equals("x-amz-request-id", StringComparison.CurrentCultureIgnoreCase))
-                errorResponse.RequestId = parameter.Value;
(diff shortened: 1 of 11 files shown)


Minio/MinioClient.cs
```diff
@@ -646,17 +646,14 @@ private static void ParseWellKnownErrorNoContent(ResponseResult response)
         MinioException error = null;
         var errorResponse = new ErrorResponse();
 
-        foreach (var parameter in response.Headers)
-        {
-            if (parameter.Key.Equals("x-amz-id-2", StringComparison.CurrentCultureIgnoreCase))
-                errorResponse.HostId = parameter.Value;
+        if (response.Headers.TryGetValue("x-amz-id-2", out string hostId))
```
pbolduc, line 649: Note the Headers property uses `OrdinalIgnoreCase` and not `CurrentCultureIgnoreCase`. I wouldnt expect header keys to have locale specific encodings. The values might?

PR #742, opened 2023-02-03
clean up HttpRequestMessageBuilder add query parameters

pbolduc:
This PR cleans up some of the `HttpRequestMessageBuilder` methods that add query parameters. String interpolation is significantly slower than using a `ToString()` function. Take this benchmark,

```cs
using BenchmarkDotNet;
using BenchmarkDotNet.Attributes;
using BenchmarkDotNet.Jobs;
using BenchmarkDotNet.Running;
using System.IO;
using System.Text;
using System.Xml;
using System.Xml.Linq;
using System.Collections.Generic;
using System;

BenchmarkRunner.Run<Benchmarks>();

[MemoryDiagnoser]
public class Benchmarks
{
    [Params(10)]
    public int n;

    [Benchmark(Baseline = true)]
    public string ToString() => n.ToString();

    [Benchmark]
    public string StringInterpolation() => $"{n}";
}
```

shows that using string interpolation is 5.56 x slower on integers.  Also there were a number of places where string interpolation was being used on string values (`VersionId` or `UploadId`) unnecessarily.

```
BenchmarkDotNet=v0.13.4, OS=Windows 11 (10.0.22621.1105)
AMD Ryzen Threadripper PRO 3955WX 16-Cores, 1 CPU, 32 logical and 16 physical cores
.NET SDK=7.0.102
  [Host]     : .NET 6.0.13 (6.0.1322.58009), X64 RyuJIT AVX2
  DefaultJob : .NET 6.0.13 (6.0.1322.58009), X64 RyuJIT AVX2


|              Method |  n |      Mean |     Error |    StdDev | Ratio | RatioSD |   Gen0 | Allocated | Alloc Ratio |
|-------------------- |--- |----------:|----------:|----------:|------:|--------:|-------:|----------:|------------:|
|            ToString | 10 |  7.085 ns | 0.1979 ns | 0.2118 ns |  1.00 |    0.00 | 0.0038 |      32 B |        1.00 |
| StringInterpolation | 10 | 39.342 ns | 0.2033 ns | 0.1901 ns |  5.56 |    0.20 | 0.0038 |      32 B |        1.00 |
```

### DIFF
--- a/Minio/DataModel/ObjectOperationsArgs.cs
+++ b/Minio/DataModel/ObjectOperationsArgs.cs
@@ -59,7 +59,7 @@ public class SelectObjectContentArgs : EncryptionArgs<SelectObjectContentArgs>
     internal override HttpRequestMessageBuilder BuildRequest(HttpRequestMessageBuilder requestMessageBuilder)
     {
         requestMessageBuilder.AddQueryParameter("select", "");
-        requestMessageBuilder.AddQueryParameter("select-type", "2");
+        requestMessageBuilder.AddQueryParameter("select-type", 2);
 
         if (RequestBody == null)
         {
@@ -231,7 +231,7 @@ public class StatObjectArgs : ObjectConditionalQueryArgs<StatObjectArgs>
     internal override HttpRequestMessageBuilder BuildRequest(HttpRequestMessageBuilder requestMessageBuilder)
     {
         if (!string.IsNullOrEmpty(VersionId))
-            requestMessageBuilder.AddQueryParameter("versionId", $"{VersionId}");
+            requestMessageBuilder.AddQueryParameter("versionId", VersionId);
         if (Headers.ContainsKey(S3ZipExtractKey))
             requestMessageBuilder.AddQueryParameter(S3ZipExtractKey, Headers[S3ZipExtractKey]);
 
@@ -445,7 +445,7 @@ public class RemoveUploadArgs : EncryptionArgs<RemoveUploadArgs>
 
     internal override HttpRequestMessageBuilder BuildRequest(HttpRequestMessageBuilder requestMessageBuilder)
     {
-        requestMessageBuilder.AddQueryParameter("uploadId", $"{UploadId}");
+        requestMessageBuilder.AddQueryParameter("uploadId", UploadId);
         return requestMessageBuilder;
     }
 }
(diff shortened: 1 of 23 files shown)


Minio/DataModel/ObjectOperationsArgs.cs
```diff
@@ -1691,7 +1691,6 @@ internal override HttpRequestMessageBuilder BuildRequest(HttpRequestMessageBuild
                 new XElement("ETag", ETags[i])));
         var completeMultipartUploadXml = new XElement("CompleteMultipartUpload", parts);
         var bodyString = completeMultipartUploadXml.ToString();
-        var body = Encoding.UTF8.GetBytes(bodyString);
```
pbolduc, line 1694: unrelated to `HttpRequestMessageBuilder`, but this was not being used and just converting the string twice.

Minio/DataModel/ObjectOperationsArgs.cs
```diff
@@ -1712,7 +1711,7 @@ internal override void Validate()
     {
         base.Validate();
         if (string.IsNullOrWhiteSpace(UploadId))
-            throw new ArgumentNullException(nameof(UploadId) + " not assigned for PutObjectPart operation.");
+            throw new ArgumentNullException(nameof(UploadId), $"{nameof(UploadId)} not assigned for PutObjectPart operation.");
```
pbolduc, line 1714: The single argument constructor takes the argument name as the paramName. The second constructor parameter is paramName, message.

[`public ArgumentNullException (string? paramName);`](https://learn.microsoft.com/en-us/dotnet/api/system.argumentnullexception.-ctor?view=net-7.0#system-argumentnullexception-ctor(system-string)) vs

[`public ArgumentNullException (string? paramName, string? message)`](https://learn.microsoft.com/en-us/dotnet/api/system.argumentnullexception.-ctor?view=net-7.0#system-argumentnullexception-ctor(system-string-system-string))

Minio/HttpRequestMessageBuilder.cs
```diff
@@ -172,6 +172,11 @@ public void AddQueryParameter(string key, string value)
         QueryParameters[key] = value;
     }
 
+    public void AddQueryParameter(string key, int value)
```
pbolduc, line 175: Added to make it easier to add an integer value, doing `.ToString()` in one place.

pbolduc:
I will review the editorconfig violations.

PR #743, opened 2023-02-06
Response cleanup

pbolduc:
This PR cleans up the Response types.

- Removed a few places where the content string is converted back to byte array just to be read from a memory stream. They bytes are available as a property.
- Change `GenericResponse` to be abstract
- Added abstract` GenericXmlResponse<TResult>` that inherits from `GenericResponse`
   - consolidates the logic to deserialize the response content as XML of type TResult
   - has optional function to convert content for those few xml types that need namespaces removed
- Change almost all response types to only take a single `ResponseResult` constructor argument
- Fix a few`ExecuteTaskAsync` calls that did not `.ConfigureAwait(false)` on them
- Remove the `Response.Content.ReadAsStreamAsync().Result)` and replaced with `Response.Content.ReadAsStreamAsync().ConfigureAwait(false).GetAwaiter().GetResult()`
- Add `ResponseHeaderCollection` for consistent Ordinal Ignore-Case header lookup
- Add extension method on `IReadOnlyDictionary<string, string>` to get a header as string or null if it doesn't exist `GetValueOrNull`
- Added a `VirutalStream` class to the Test project for future testing of uploading large streams without really allocating memory or using disk.
- Try to clean up ResponseResult's `ContentBytes`, `Content` and `ContentStream`
   - removed ContentStream because reading the result content more than once should throw an error
   - move the ContentStream logic into the initializing `ContentBytes` lazyily.
 - Replaced ResponseResult's `public Dictionary<string, string> Headers` property with a initialized ResponseHeaderCollection based on the incoming response. If there is no response, headers gets initialized to statically allocated `ResponseHeaderCollection.Empty` instance.

### DIFF
--- a/Minio.Tests/VirtualStream.cs
+++ b/Minio.Tests/VirtualStream.cs
@@ -30,7 +30,7 @@ namespace Minio.Tests;
 /// Virtural stream will create a random data stream up to the user's specific size.
 /// It will not allocate any data. This stream can only be read forward.
 /// </summary>
-public class VirutalStream : Stream
+public class VirtualStream : Stream
 {
     /// <summary>
     /// The current position in the stream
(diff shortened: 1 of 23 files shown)


pbolduc:
I will review the editorconfig vioations.

pbolduc:
> I will review the editorconfig violations.

There is no .editorconfig file in the solution that I can find. It is difficult to fix these issues without this file. Can you point me in the right direction, otherwise it will be impossible for me to resolve.

PR #749, opened 2023-02-20
Replaces 'http' with 'https' in Minio.Tests for 'play.min.io'

ebozduman:
Replaces Minio client `endpoint` `http://play.min.io:80` with `https://play.min.io:443` for `Minio.Tests`.
The `Minio.Tests` started failing during build tests after `play` server is decommissioned.
Anyways, we were supposed to use `https`.


ebozduman:
Removed the port number `443` as it is the default, per @kannappanr's review comment.

kannappanr (approved):

PR #732, opened 2023-01-13
add .Net6 support across codebase

martijn00:
This removes dependencies on old libraries: fixes #725

PR #754, opened 2023-03-06
Use span and Hashing for Crc32

martijn00:
To remove the dependency on Crc32.NET and make use of Span for efficiënt memory usage.

PR #756, opened 2023-03-23
Netstandard compat

martijn00:
I still don't think this is a good idea, but anyways, here it is.

See: https://github.com/minio/minio-dotnet/pull/732#issuecomment-1481056375


martijn00:
@ebozduman if you want to continue on this path, please review and merge. After that I can also update my Span PR: https://github.com/minio/minio-dotnet/pull/754

ebozduman:
Reviewing...

PR #763, opened 2023-04-13
Deserialize xml content not containing namespace

martijn00:
Followup on https://github.com/minio/minio-dotnet/pull/762


ebozduman (changes requested):
Please notice there are 3 github actions reported for:
1- The variable 'ex' is declared but never used
2- The annotation for nullable reference types should only be used in code within a '#nullable' annotations context
3- XML comment is not placed on a valid language element

ebozduman (approved):
LGTM

PR #786, opened 2023-05-19
PutObjectAsync with Snowball header must not use multipart

emrocha:
PutObjectAsync must not use multipart if has header "X-Amz-Meta-Snowball-Auto-Extract=true".
See https://github.com/minio/minio/issues/17033


ebozduman (commented):
LGTM
But we should keep in mind that this is not enough to support Snowball.

emrocha:
Yes. I could write a function similar to upload_snowball_objects from the Python client.

ebozduman (approved):
LGTM (#797 addresses the missing snowball support)

PR #796, opened 2023-06-19
Warnings again

martijn00:
@ebozduman lots of file movements here to better structure the code and remove warnings. Can you help and add to this PR with fixes for the warnings? After this gets merged the analyzer can be enabled by default!


ebozduman:
@martijn00 ,

The lint changes are ready to be pushed but before doing that I noticed there are total `363` files, all marked as `rewrite` by the `git commit` command. However this PR shows only `278` files as being modified. I think we need to understand this before pushing the changes.

    ..... after a while ......

OK. I think this is because you work in a Windows environment and even the unmodified files come with a change since every lilne ends with `CRLF`, which are automatically replaced with `LF` by git. Although it is kind of late as they became part of the PR. This should have been taken care of before the PR is created.

So, please make sure the end of line character in your VS Code settings, `Files:Eol`, is set to `\n`.

martijn00:
@ebozduman This branch now builds and has the lowest warning setting enabled. It is ready to be merged and in new PR's I can set the warnings stricter.

martijn00:
@ebozduman Can you run CI again?

ebozduman (approved):
LGTM
(Too many files make the review process very difficult and increases the chance of some unwanted changes to slip through inadvertently.)

PR #799, opened 2023-06-29
Fixes .net4.6~.net4.71 framework support issue for ssl protocol setting

ebozduman:
Subnet cusomer hit the error `Operation not supported in this platform` while creating a certificate identity provider in `net462`.

That is because `System.Net.Http.HttpClientHandler` did not support `SslProtocols` property to set the protocol type for frameworks before `net472` release. 

So the protocol type is needed to be set using `System.Net.ServicePointManager.SecurityProtocol` property to support older frameworks like `net462`.


martijn00 (approved):

martijn00:
@ebozduman can you rebase this one?

martijn00:
@ebozduman it looks like something went wrong? Did you merge instead of force rebase?

ebozduman:
I still see the CRLF issue in my branch and I was trying to fix it. 

The problem is `dotnet regitlint` check hits the CRLF => LF issue and not proceeding further during build test.

I think we need to revert the PR  #798 to go back to the old .gitattribute setting.
However, `text=auto` does not help.

I expect this forced commit will be cleaned up when the CRLF => LF replacement PR #802, I've just created from `crlf_cleanup` branch is merged into this PR.

ebozduman:
@martijn00 

I agree. I'll create a new PR and close this one.

ebozduman:
Created a new PR, #836, which supersedes this one.
So, this one is closed.

PR #802, opened 2023-07-10
eol normalization with CRLF => LF in all files

ebozduman:
Replaces all CRLF with LF character in all files.
This cleanup is required as a first step to clean up the master.


ebozduman:
@martijn00 

lint complains about the mismatch between the following files and .editorconfig.
I guess this is due to the .gitattribute change this PR brings, `text=lf` to resolve CRLF => LF issue.

```
!!!! Process Aborted !!!!
The following files do not match .editorconfig:
 * Directory.Build.props
 * FileUploader/FileUpload.cs
 * Minio.Examples/Cases/AWSEnvironmentProviderExample.cs
 *  ... and all Minio.Examples/Cases files
 : 
 :
```

Can we relax .editorconfig to let these files pass the lint check so that we can replace CRLF with LF eol character?Later, we can put them all back together with .gitattribute setting `text=auto`.

martijn00:
I'm not sure what you plan to do... Why not just force push the correct line endings to master?

Here you can find the editorconfig line: https://github.com/minio/minio-dotnet/blob/master/.editorconfig#L11

ebozduman:
> I'm not sure what you plan to do... Why not just force push the correct line endings to master?

@martijn00 

That's exactly what I am trying to do. 
The issue with "text=auto` is the fact that the `git add` command after `dotnet regitlint` unmarks the modified files and accepts the files with `crlf` chars quietly and causing the lint check hit the crlf problem during the build tests.
So, this PR is equivalent to forcing `lf` for eol character in all files.

martijn00:
But why revert my PR then? Why not just force it into the master branch?

ebozduman:
> But why revert my PR then? Why not just force it into the master branch?

@martijn00 

You are right. That is wrong in the title and in the description.
So, I did not revert the PR798.
The only change in the .gitattribute file is this:
```
* text=auto eol=lf  =>  * text=lf
```
I will remove the references to PR798 in the title and in the description.

krisis (approved):

harshavardhana (approved):

Issue #808, opened 2023-07-26
ObjectStat.ObjectLockRetainUntilDate contains wrong value


ante-maric:
I believe there is bug in this code here:

```
case "x-amz-object-lock-retain-until-date":
                    var lockUntilDate = paramValue;
                    if (!string.IsNullOrWhiteSpace(lockUntilDate))
                        objInfo.ObjectLockRetainUntilDate = DateTime.SpecifyKind(
                            DateTime.Parse(lockUntilDate, CultureInfo.CurrentCulture),
                            DateTimeKind.Utc);
```

DateTime.Parse will convert lockUntilDate to local time. Then that local time changes kind to Utc. So, if your time zone is +2h what you will get back is 2h more than you set.

martijn00:
Can you make a PR to fix this?

ante-maric:
> Can you make a PR to fix this?

I'll try to do it this week. There might be some other places where DateTime.Parse is being used, I'll check those too.

ebozduman:
@ante-maric,

Could you test your scenario with #811 fix?

PR #811, opened 2023-07-31
Fixed time zone lost when serializing and parsing dates

### LINKED ISSUE #810: RetentionUntilDate in SetObjectRetentionArgs serialized to wrong value
When RetentionUntilDate property in SetObjectRetentionArgs class serialized, it formatted to ISO 8601 string, but time zone offset get lost. 
```
public static string To8601String(DateTime dt)
{
    return dt.ToString("yyyy-MM-dd'T'HH:mm:ssZ", CultureInfo.InvariantCulture);
}
```
As result if date in local time zone, minio server received date in past or future depending on time zone.
I tried pass date in UTC, but get error, because passed value checked against local time (DateTime.Now) always

```
if (DateTime.Compare(RetentionUntilDate, DateTime.Now) <= 0)
            throw new InvalidOperationException("Retention until date set using " + nameof(WithRetentionUntilDate) +
                                                " needs to be in the future.");
```

basrach:
Fixes #810 and #808


ante-maric:
I would suggest that the whole library should work only with UTC and that names of parameters and properties clearly state that by using UTC suffix which is common convention. That would greatly simplify things. Working with local time bites you sooner or later.

martijn00:
@ebozduman this looks good to go to me

martijn00:
@basrach Try to do a Rebase instead of merge to fix the code style.

ebozduman (approved):
LGTM

PR #814, opened 2023-08-01
Rework the github workflow

martijn00:
@ebozduman 

According to good practice and latest standards there are some things that should change. You need to create a develop and a main branch. The default branch should be develop.

Also a NUGET_API_KEY needs to be added to the secrets


martijn00:
@ebozduman can you check this and run?

martijn00:
@ebozduman can you run again?

ebozduman:
@martijn00 

It is stuck again !?!? Is the added master branch gone somehow?

Meanwhile,
Not now but later, we can also add one of the old officially supported .net framework versions by MicroSoft , 
```
4.6.1, 4.6.2, 4.7, 4.7.1, 4.7.2, 4.8, 4.8.1
```
let's say  `4.7.2` or `4.8.1`, as a legacy net framework sample build check.

martijn00:
Yeah we can add a framework like that later. The build is not stuck, you just need to run the last commit.

ebozduman:
Started.
It did not display the check `Start` button first, but now I see it.

martijn00:
Look at the flow now: https://github.com/minio/minio-dotnet/actions/runs/5738231383

Looks nice with jobs splitting up! Later on I could maybe even bring in more improvements.

martijn00:
@ebozduman it passed the build! Can you merge now?

ebozduman:
Running again

ebozduman:
By the way, I also noticed the entry `end_of_line = crlf` rule in .editorConfig file.
That would trip your PR #815 check.
I am pretty sure, I replaced it with `lf`, but I don't see any records of it nowhere in github, and  #802 also did not have it.
I'll create a PR for it.

ebozduman (approved):
LGTM

ebozduman:
@martijn00 

The build check is waiting for status to be reported for a 6.0 build,  which is supposed to be not required anymore.

martijn00:
How can that be removed? It is not in the new script anymore

ebozduman:
I am also not sure how. Checking...

martijn00:
@ebozduman I found this:

The status "Expected — Waiting for status to be reported" typically appears when a GitHub Actions workflow is expected to run on a commit or pull request, but has not yet reported a status. If the workflow was removed or renamed in the repository, this status might be left "hanging" and the check might become a "stale check."

To remove a stale check, you need to update the branch protection rules, if you have the necessary permissions (i.e., you are an owner or administrator of the repository). Here's how to do it:

Go to your GitHub repository.
Click on "Settings" on the right side of the repository's page.
Click on "Branches" in the left sidebar.
Under the "Branch protection rules" section, click on the "Edit" button next to the rule that applies to the branch you're concerned about.
In the "Protect matching branches" section, scroll down to the "Require status checks to pass before merging" area.
You should see a list of status checks found in the last week for this repository. If the stale check is listed here, uncheck it.
Click on "Save changes" at the bottom of the page.
If the stale check isn't listed, it could be that it's not a check from the last week. In this case, you may have to temporarily re-add the check (for example, by restoring the workflow that produced it), push a commit to trigger it, then remove it from your branch protection rules.

Note that only users with admin or owner permissions for the repository can change the branch protection rules. If you don't have the necessary permissions, you'll need to contact someone who does to make the change for you.

PR #812, opened 2023-08-01
Rework parallel util method

martijn00:
@ebozduman what about this one

PR #815, opened 2023-08-01
Normalize all the line endings

martijn00:
@ebozduman this seems to finally have done it. Can you merge?


martijn00:
@ebozduman I'm trying to figure this out again. Yesterday it worked for me to convert all using `find . -type f -exec dos2unix {} \;    ` today it seems all reverted back.

I'm struggling to move forward because this process takes so long and so many PR's need updating every time. Can you merge #814 and create the develop and main branch? Then it is easier to work forward on next PR's.

ebozduman:
@martijn00 

I'll check PR#814, but I think we need #812 merged to be able to turn off parallel run. Only then we can start seeing successful build tests and merges happening.

martijn00:
@ebozduman Then lets merge 812 and 814 right after. I still doubt if the build runs with formatting on that. But lets move forward first and merge all those PR's and fix the formatting after in a new PR.

PR #832, opened 2023-08-09
Reverses the wrong condition when a new bucket cache is needed

ebozduman:
This faulty code was there for quite some time.
The issue is fixed by reversing the faulty if condition when there is no BucketCache class is instantiated.

This fix addresses the build test failure hit during PR #830 functional testing.


martijn00 (approved):

ebozduman:
@martijn00 ,

I had the functional tests run without hitting any problem in my environment.
But when I pushed it, the same functional test fails.
Currently looking into it.

ebozduman:
It is ready for review now.

harshavardhana (approved):

PR #836, opened 2023-08-11
Sets sslProtocol per dotnet version

ebozduman:
This PR supersedes #799 (can be closed).

Subnet customer hit the error "Operation not supported", while trying to create a certificate identity provider using an old but still supported dotnet release version, .net462.

That is because `System.Net.Http.HttpClientHandler` did not support `SslProtocols` property to set the protocol type for all supported dotnet release versions older than .net472.

So the protocol type is needed to be set using `System.Net.ServicePointManager.SecurityProtocol` property to support dotnet release versions older than .net472.

### DIFF
--- a/Minio/Credentials/CertificateIdentityProvider.cs
+++ b/Minio/Credentials/CertificateIdentityProvider.cs
@@ -16,7 +16,6 @@
  */
 
 using System.Globalization;
-using System.Security.Authentication;
 using System.Security.Cryptography.X509Certificates;
 using System.Text;
 using System.Web;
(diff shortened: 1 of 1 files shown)


martijn00 (approved):

martijn00:
@ebozduman @harshavardhana can you merge this?

ebozduman:
@harshavardhana ,
I cannot merge this only with @martijn00's approval

ebozduman:
@praveen, @balamurugana , @krisis 
Just need 1 review approval.
Thanks

balamurugana (approved):

PR #850, opened 2023-08-25
Fix inheritance and add servicecollection

martijn00:
@ebozduman


ebozduman:
@martijn00  ,

Will do the review in during CA work hours...

ebozduman (commented):
@martijn00 

Could you add some explanation/information about this PR, like what inheritance issue is fixed by this PR and why service collection is added, what is it going to help/solve etc, briefly in just one or two sentences?

martijn00:
The IMinioClient interface didn't inherit the 2 other interfaces, but the implementation did implement them. For that reason the methods in the interface were also duplicated. I've removed that.

I'm now working towards a more interface based implementation of minio so that it the standard DI container of .net can use it. 

This is a first step towards an easier way to implement Minio. You can call AddMinio in your normal api build and resolve IMinioClient or IMinioClientFactory through DI or the service locator after that

ebozduman (approved):
LGTM

PR #952, opened 2023-12-12
Increase Developer Experience by throwing an exception on MemoryStream's zero position logical error.

tayyebi:
There will be a logical error in case of  reading from MemoryStream. A condition is added, in case which developer really decided to iterate over a memory stream (Even it is better to use a `prop` in this case).

This will handle scenarios such as below:
```
public async Task UploadFileAsync(IFormFile f)
{
  using (var ms = new MemoryStream())
  {
    f.CopyTo(ms);
    // !!! --------> ms.Position is equal to it's Length here

```


ebozduman:
@tayyebi 
I am confused and not sure what you are trying to achieve.
There are typos, and errors in the code, so the build is going to fail.

ebozduman (changes requested):
Your code needs to be fixed. There are problems like:
- `ObjectStreamData.Lenght` : 
`ObjectStreamData` is not defined. You meant to use `args.ObjectStreamData`, I guess.
`Lenght` has a typo...

Please make sure you test your code by making a build, running tests etc. before creating a PR.
I suggest also you to use some IDE environment which helps these kind of errors while you work on your code.

ebozduman (commented):
Besides, this part of the code is not designed and implemented to take care of a file with multiple parts.
That's why `ReadFullAsync` method is called to read full data.

Please notice the name of the method returned, `PutObjectSinglePartAsync` at the end of this block.

If your goal is to send a file < 5MG in multi-parts, that is not something we are interested in or support.

So, this PR, as is today, is rejected.

If you disagree, please feel free to explain your goal and tell us what Amazon s3 feature/requirement is going to be met.

ebozduman:
The reasoning is explained above.

tayyebi:
Sorry for bothering you. Let me explain a scenario:
Let's suppose an action which accepts a file from user and directly uploads it to MinIO.
```
public async Task UploadFileAsync(IFormFile image)
{
  using (var ms = new MemoryStream())
  {
    image.CopyTo(ms);
    args = new Minio.DataModel.Args.PutObjectArgs { }
    .WithBucket(bucket)
    .WithObject(Guid.NewGuid().ToString())
    .WithContentType(image.ContentType)
    .WithStreamData(ms)
    .WithObjectSize(ms.Length);
    await minioClient.PutObjectAsync(args);
  }
}
```

The function `CopyTo` will increase the Position (which is a pointer) of a `MemoryStream`, regarding to the source code of _MinIO_, it will not reset the `MemoryStream` instance's pointers (which is somehow a good decision if there are multiple files stored in a `MemoryStream` and someone wants to break it in chunks. So, a `ms.Position = 0;` is required before passing it to MinIO SDK.

But I think it's better to throw an exception to notify developer about the issue: `if ((int)args.ObjectSize !=  args.ObjectStreamData.Lenght - args.ObjectStreamData.Position)`

What do you think?

I can send another PR, if this is a good idea.

Cheers.

PR #955, opened 2023-12-13
re-adjusted the progress reporting

ebozduman:
Sometimes the build test fails when 100% progress is supposed to be reached at for `PutObject_Test9` and `PutObject_Test10` tests.
So, the progress code is re-adjusted to address this sporadic issue.
Since I cannot reproduce the problem and hence cannot test the fix in my local environment, it would not be possible to say for sure the issue is addressed even if the build tests pass.

### DIFF
--- a/Minio.Functional.Tests/FunctionalTest.cs
+++ b/Minio.Functional.Tests/FunctionalTest.cs
@@ -3494,13 +3494,19 @@ public static class FunctionalTest
         var startTime = DateTime.Now;
         var bucketName = GetRandomName(15);
         var objectName = GetRandomObjectName(10);
-        var contentType = "application/octet-stream";
+        const int objSize = 1 * MB;
+        const string contentType = "application/octet-stream";
         var percentage = 0;
         var totalBytesTransferred = 0L;
         var progress = new Progress<ProgressReport>(progressReport =>
         {
             percentage = progressReport.Percentage;
             totalBytesTransferred = progressReport.TotalBytesTransferred;
+            // Console.WriteLine(
+            //    $"PutObject_Test9 - Percentage: {progressReport.Percentage}% TotalBytesTransferred: {progressReport.TotalBytesTransferred} bytes");
+            // if (progressReport.Percentage != 100)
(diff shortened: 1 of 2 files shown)


Minio.Functional.Tests/FunctionalTest.cs
```diff
@@ -3494,13 +3494,19 @@ internal static async Task PutObject_Test9(IMinioClient minio)
...
-        var contentType = "application/octet-stream";
+        const int objSize = 1 * MB;
+        const string contentType = "application/octet-stream";
         var percentage = 0;
         var totalBytesTransferred = 0L;
         var progress = new Progress<ProgressReport>(progressReport =>
         {
             percentage = progressReport.Percentage;
             totalBytesTransferred = progressReport.TotalBytesTransferred;
+            // Console.WriteLine(
```
shtripat, line 3505: Do we need these lines for reference?

Minio.Functional.Tests/FunctionalTest.cs
```diff
@@ -3548,11 +3554,11 @@ internal static async Task PutObject_Test10(IMinioClient minio)
         {
             percentage = progressReport.Percentage;
             totalBytesTransferred = progressReport.TotalBytesTransferred;
-            //Console.WriteLine(
-            //    $"Percentage: {progressReport.Percentage}% TotalBytesTransferred: {progressReport.TotalBytesTransferred} bytes");
-            //if (progressReport.Percentage != 100)
+            // Console.WriteLine(
```
shtripat, line 3557: Same as above

shtripat (approved):
Minor comment/clarification.

Minio.Functional.Tests/FunctionalTest.cs
ebozduman: I left those as an example on how to print the progress information.
We don't want to introduce random print outs in between functional test results.

@shtripat 
Please review again as I've changed the way the size of the put object info is retrieved when the file is uploaded as a single file in one shot.

shtripat (approved):
lgtm

PR #956, opened 2023-12-15
adds the missing NullForgiving operator comment

ebozduman:
Bumping `Nullable.Extended.Analyzer` from version `1.10.4539` to `1.14.6129` caused the build process to fail. 
That is because the Analyzer checks the usage of `NullForgiving operator` and require a `justification` for each use case. If it cannot find the justification, then it throws an exception during the build process.
The justification is a simple comment next the code and it is supposed to start with `// ! ` and gives some explanation about the use case.


shtripat (approved):
lgtm

PR #974, opened 2024-01-07
Fix analyzer update

martijn00:


Minio/DataModel/Args/GetObjectArgs.cs
```diff
@@ -76,8 +75,7 @@ internal override HttpRequestMessageBuilder BuildRequest(HttpRequestMessageBuild
     {
         if (!string.IsNullOrEmpty(VersionId)) requestMessageBuilder.AddQueryParameter("versionId", $"{VersionId}");
 
-        if (CallBack is not null) requestMessageBuilder.ResponseWriter = CallBack;
-        else requestMessageBuilder.FunctionResponseWriter = FuncCallBack;
+        requestMessageBuilder.ResponseWriter = CallBack;
```
ebozduman, line 78: what happens if `Callback` is null?

ebozduman (approved):

PR #985, opened 2024-01-18
changes MinioClient with IMinioClient in doc examples

ebozduman:
API.md still had the old `MinioClient` instead of `IMinioClient` when a client instance is created. Replaced them all


prakashsvmx (approved):
Thank you for the update :+1:

PR #1049, opened 2024-04-09
Use IAsyncEnumerable for ListObjectsAsync instead of observables.

ramondeklein:
Using `IAsyncEnumerable` is much easier to work with compared to `IObservable`, so I have rewritten the `ListObjectsAsync` method. This also fixes a bug that causes `Minio.Examples.Cases.ListObjects.Run` to return before it has actually completed. The code is also much simpler and more efficient.

Using `IAsyncEnumerable` has a significant advantage over `IEnumerable`, because it can be cancelled and doesn't need to finish until all objects have been loaded. It actually hides the paging-mechanism to the caller, but under the hood it's still being used.

To avoid breaking changes, I have used a new name for the new method and created an extension method (with the old name) that converts the `IAsyncEnumerable` back to `IObservable` (zero cost), but marked it as obsolete to encourage clients to use the updated method.


Minio.Examples/Cases/ListObjects.cs
```diff
@@ -33,12 +33,11 @@ internal static class ListObjects
             var listArgs = new ListObjectsArgs()
                 .WithBucket(bucketName)
                 .WithPrefix(prefix)
-                .WithRecursive(recursive);
-            var observable = minio.ListObjectsAsync(listArgs);
-            var subscription = observable.Subscribe(
-                item => Console.WriteLine($"Object: {item.Key}"),
-                ex => Console.WriteLine($"OnError: {ex}"),
-                () => Console.WriteLine($"Listed all objects in bucket {bucketName}\n"));
```
ramondeklein, line 41: This code had several bugs:
1. The method returns directly after subscribing, so the events are emitted when the function has already completed.
2. The `subscription` is never disposed (but if used `using var subscription == ...`, then the subscription would be cancelled when it got out-of-scope (directly after subscribing).

Minio/ApiEndpoints/BucketOperationsExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+using Minio.DataModel;
+using Minio.DataModel.Args;
+
+namespace Minio.ApiEndpoints;
+
+public static class BucketOperationsExtensions
+{
+    // Used for backwards compatibility
+    [Obsolete("Use the IAsyncEnumerable version instead")]
+    public static IObservable<Item> ListObjectsAsync(this IBucketOperations self, ListObjectsArgs args)
```
ramondeklein, line 11: This method was added to provide compatibility with clients that rely on `IObservable`.

Minio.Examples/Program.cs
```diff
@@ -158,7 +158,7 @@ public static async Task Main()
         await StatObject.Run(minioClient, bucketName, objectName).ConfigureAwait(false);
 
         // List the objects in the new bucket
-        ListObjects.Run(minioClient, bucketName);
+        await ListObjects.Run(minioClient, bucketName).ConfigureAwait(false);
```
ramondeklein, line 161: This method properly waits until it has completed.

Minio/ApiEndpoints/BucketOperations.cs
```diff
@@ -201,62 +204,64 @@ public async Task SetVersioningAsync(SetVersioningArgs args, CancellationToken c
...
+            cancellationToken.ThrowIfCancellationRequested();
+            
+            var requestMessageBuilder = await this.CreateRequest(goArgs).ConfigureAwait(false);
+            using var responseResult = await this.ExecuteTaskAsync(ResponseErrorHandlers, requestMessageBuilder, cancellationToken: cancellationToken) .ConfigureAwait(false);
+
+            if (responseResult.StatusCode != HttpStatusCode.OK)
+                throw new ErrorResponseException($"HTTP status-code {responseResult.StatusCode:D}: {responseResult.StatusCode}", responseResult);
+
+#if NET2_0_OR_GREATER
+            var root = await XDocument.LoadAsync(responseResult.ContentStream, LoadOptions.None, cancellationToken).ConfigureAwait(false);
```
ramondeklein, line 237: Make use of async XML loading in .NET core (not available in legacy .NET Framework). Also note that XML reading is done directly on the stream and not using an intermediate string.

Minio/ApiEndpoints/BucketOperations.cs
```diff
@@ -201,62 +204,64 @@ public async Task SetVersioningAsync(SetVersioningArgs args, CancellationToken c
...
+#else
+            var root = XDocument.Load(responseResult.ContentStream);
+#endif
+
+            var items = from c in root.Root.Descendants(tag)
```
ramondeklein, line 242: Reading objects or versions is very similar, so only differentiate by tag.

Minio/ApiEndpoints/BucketOperations.cs
```diff
@@ -201,62 +204,64 @@ public async Task SetVersioningAsync(SetVersioningArgs args, CancellationToken c
...
-                    cts.Token.ThrowIfCancellationRequested();
-                    count++;
-                }
-            }
-        );
+                    Key = c.Element(ns + "Key").Value,
+                    LastModified = c.Element(ns + "LastModified").Value,
+                    ETag = c.Element(ns + "ETag").Value,
+                    Size = ulong.Parse(c.Element(ns + "Size").Value, CultureInfo.InvariantCulture),
+                    VersionId = c.Element(ns + "VersionId")?.Value ?? string.Empty,
```
ramondeklein, line 249: `VersionId` is only used when reading versions, so I set it to `string.Empty` instead if not present (could also be `null`).

martijn00:
@ramondeklein this is great! I made several attempts at IAsyncEnumerable in the past already. It would be great to remove the dependency on Reactive since it is not maintained that much.

ramondeklein:
> @ramondeklein this is great! I made several attempts at IAsyncEnumerable in the past already. It would be great to remove the dependency on Reactive since it is not maintained that much.

I'm working on a proof-of-concept to rewrite the Minio SDK for .NET, but not sure if we are going through with it.

@martijn00  Enjoy your "kapsalon" and say ✋🏻 to Thomas. Greetings from Enschede...

ebozduman (commented):
LGTM.

Btw, did you run `Example1` and/or the other tests ?
If yes, could you add some simple info about them in the PR, like how you ran the tests and the expected results, etc.

ebozduman (dismissed):
LGTM
Marked `Approve`.

ramondeklein:
@ebozduman I see that I need to fix some tests in the functional tests. Will do that later this week.

Minio.Examples/Cases/ListObjects.cs
ebozduman: Yes. It is important that we keep supporting old .NET Framework versions.
I'll start working on it and look into and fix if any problems.

martijn00: I would drop anything below .NET 6: https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core
It is worth supporting long term support like that, and .NET 4.8.x also has LTS, but is compatible with much newer APIs.

There is not many valid reasons anyone would continue to use .netstandard2.x.

martijn00:
@ramondeklein this is probably not the right place, but if you want to pick up something i haven't done yet in all the refactoring: https://github.com/minio/minio-dotnet/blob/master/Directory.Build.props#L21

Enable nullable and fix all the issues. This would help with quite some bugs.

Minio.Examples/Cases/ListObjects.cs
ramondeklein: I think we should keep this SDK compatible with .NET Framework and move it into maintenance mode (so only bugfixing and security fixes). For new functionality, I think for we should rewrite the SDK and focus on .NET 6 (and later). Rewriting the SDK is less work compared to getting this SDK up to modern .NET standards. Modern .NET requires support for the extension libraries (DI, logging, configuration, ...).

My [PoC](https://github.com/ramondeklein/minio-dotnet2) supports .NET Core 6.0 and up. It allows for much more scalable software (more async functions). I don't think anyone started a .NET Framework project the past 3-5 years anymore, so there is no need to support .NET Framework for new developments.

I think I'll drop the unit-tests for Minio functionality from the PoC, because the integration tests are easier to maintain and also guarantee proper functionality. Only unit tests for basic things (like AWS V4 authentication) seem legit.

ramondeklein: @martijn00 Developing a new .NET SDK won't have the highest priority (if it's decided to even continue to work on it), so I think .NET 9 will be around when the new SDK released. But I currently did implement shims for missing .NET 6 functionality. It is recommended to use .NET 7, because it won't use shims.

ramondeklein:
@ebozduman I also added the user metadata changes from #1043 to this PR. There is a compatibility extension, so code that is based on `IObservable<>` should still work.

ramondeklein:
The updated version also doesn't need any delays anymore (I've removed them from the tests) and all tests are still working.

harshavardhana (approved):

ebozduman:
I just finished running my verification tests and my review.
I was going to approve the PR, but
@harshavardhana beats again.

delasource:
when will this be released? The official quick start documentation links to a file changed in this PR, but that code (i.e. the newly named function ListObjectsEnumAsync) is not available in the latest release version.

Aka the documentation is in a wrong state until this releases.

ramondeklein:
@ebozduman Can we release the latest `master` to NuGet?

ebozduman:
Yes, we can as soon as the regression issue, PR#1106, is resolved.
I am looking into your review comments. I will implement/fix according to your comments and ask your review again.
Then, we can merge and create a new nuget package.

Issue #1027, opened 2024-03-03
Minio 6.0.2  - GetObjectAsync doesn't working properly with .NET 8 


dawcza3:
Hi,

I have a problem after migration to latest minio version (from 5.0.0 to 6.0.2). I'm trying to download a file using GetObjectAsync in following way:

```
 byte[] fileResult = null;
 var args = new GetObjectArgs()
                .WithBucket(bucketName)
                .WithFile(fileName)
                .WithObject(fileName)
                .WithCallbackStream(stream =>
                {
                    byte[] buffer = new byte[16 * 1024];
                    using (MemoryStream ms = new MemoryStream())
                    {
                        int read;
                        while ((read = stream.Read(buffer, 0, buffer.Length)) > 0)
                        {
                            ms.Write(buffer, 0, read);
                        }
                        fileResult = ms.ToArray();
                    }
                });

            var result = await minioClient.GetObjectAsync(args, cancellationToken);

            return fileResult;       
```

but it always return information that file have 0 size. Did something change from version 5 - do I have to construct args in different way to find and download my existing files ?

holterbades:
Probably you encounter an exception while downloading the object from the object-store. It's probably the same issue described in #1041. If you try the version 6.0.1 this error will probably vanish.

Issue #1020, opened 2024-02-23
No exception when getting objects with invalid credentials


csireszabolcs:
### Steps To Reproduce
Add and configure MinIo client with invalid credentials (AccessKey or SecretKey)

Call GetObjectAsync:
`await minioClient.GetObjectAsync(args);`

No exception throws, returns ObjectStat with empty values

### Expected behavior
Return AccessDeniedException
Working properly with version 6.0.1

### Version
Minio 6.0.2

ebozduman:
Hi @csireszabolcs,

You are right. It needs to be fixed.

holterbades:
It's the same issue described in #1041 with an active PullRequest #1057.

ebozduman:
This is the same issue with #1121 and fixed by PR#1320

PR #1057, opened 2024-04-18
Fixed error handling in `RequestExtensions`

### LINKED ISSUE #1041: Minio 6.0.2: Unexpected Behavior with Unavailable Endpoints
**Issue Summary:**

Upgrading from Minio .NET SDK version 6.0.1 to 6.0.2 has introduced unexpected behavior when interacting with non-existent Minio server endpoints. Specifically, methods like `BucketExistsAsync` and `GetObjectAsync` do not behave as expected when provided with endpoints that are not running Minio servers. Instead of throwing exceptions or indicating errors, they return unexpected results, such as always returning true for bucket existence checks and returning empty streams for object downloads.

**Steps to Reproduce:**

1. Use Minio .NET SDK version 6.0.2.
2. Configure the SDK with an endpoint that does not host a Minio server (e.g., `"127.0.0.1:11111"`).
3. Use the `BucketExistsAsync` method to check for the existence of a bucket that doesn't exist.
4. Use the `GetObjectAsync` method to download an object from a non-existent bucket.

```csharp
using Minio;

var s3Client = new MinioClient()
    .WithEndpoint("127.0.0.1:11111")
    .WithCredentials("accesskey", "secretkey")
    .Build();

//Should throw an exception, but it does not in 6.0.2, in 6.0.1 it throws a System.AggregateException
//Instead, in 6.0.2 it just returns true even though the Minio server does not exist, nor the bucket.
bool bucketExists = s3Client.BucketExistsAsync(new Minio.DataModel.Args.BucketExistsArgs().WithBucket("bucketname")).Result;

if (bucketExists)
{
    Console.WriteLine("Bucket exists according to Minio");
}

MemoryStream stream = new MemoryStream();

// An exception should be thrown here, but it is not, instead, it returns an empty stream even though neither the server, bucket, or object exists.
await s3Client.GetObjectAsync(new Minio.DataModel.Args.GetObjectArgs().WithBucket("bucketname").WithObject("objectname").WithCallbackStream(x => x.CopyTo(stream)));

Console.WriteLine($"Stream length: {stream.Length}");
```

**Expected Behavior:**

- The BucketExistsAsync method should throw an exception or return false when provided with an endpoint that does not host a Minio server.
- The GetObjectAsync method should throw an exception when attempting to download an object from a non-existent bucket.

**Actual Behavior:**

- When using BucketExistsAsync with a non-existent Minio server endpoint, it consistently returns true, incorrectly indicating the existence of the bucket.
- When using GetObjectAsync with a non-existent bucket or object, it returns an empty stream instead of throwing an exception or indicating an error.

**Environment Details:**

- Minio .NET SDK version: 6.0.2
- Operating System: Windows 11
- .NET 8

holterbades:
Response error handling was unintentionally removed in a previous commit. Added it back to the `ExecuteTaskCoreAsync` method. Fixed `NullReferenceException`-bug in the `HandleIfErrorResponse` method - `throw response.Exception;` does not make sense if the exception is null.

Fixes #1041 , #1020 , and probably #1027

### DIFF
--- a/Minio.Functional.Tests/FunctionalTest.cs
+++ b/Minio.Functional.Tests/FunctionalTest.cs
@@ -1,4 +1,4 @@
-﻿/*
+/*
  * MinIO .NET Library for Amazon S3 Compatible Cloud Storage,
  * (C) 2017-2021 MinIO, Inc.
  *
@@ -4139,9 +4139,9 @@ public static class FunctionalTest
             }
             catch (Exception ex)
             {
-                Assert.AreEqual(
+                Assert.IsTrue(ex.Message.StartsWith(
                     "MinIO API responded with message=At least one of the pre-conditions you specified did not hold",
-                    ex.Message);
+                    StringComparison.InvariantCulture));
             }
 
             new MintLogger("CopyObject_Test7", copyObjectSignature,
--- a/Minio.Functional.Tests/FunctionalTest.cs
+++ b/Minio.Functional.Tests/FunctionalTest.cs
@@ -365,9 +365,7 @@ public static class FunctionalTest
         try
         {
             if (await minio.BucketExistsAsync(beArgs).ConfigureAwait(false))
-            {
                 await minio.RemoveBucketAsync(rbArgs).ConfigureAwait(false);
-            }
 
             var found = await minio.BucketExistsAsync(beArgs).ConfigureAwait(false);
             Assert.IsFalse(found);
@@ -3527,8 +3525,8 @@ public static class FunctionalTest
             var stream = rsg.GenerateStreamFromSeed(objSize);
             var statObj = await PutObject_Tester(minio, bucketName, objectName, null, contentType, 0, null,
                 stream, progress).ConfigureAwait(false);
-            Assert.IsTrue(percentage == 100);
-            Assert.IsTrue(totalBytesTransferred == objSize);
+            Assert.IsTrue(percentage == 100, "Reported percentage after finished upload was not 100 percent.");
+            Assert.IsTrue(totalBytesTransferred == objSize, "Transfered object size does not match with the original object size.");
             new MintLogger(nameof(PutObject_Test9), putObjectSignature,
                 "Tests whether PutObject with progress passes for small object", TestStatus.PASS,
                 DateTime.Now - startTime,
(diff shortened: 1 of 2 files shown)


holterbades:
To the reviewer: I also noticed a minor codesmell: In line 110 and 113 there are null-checks to the `path` variable, which are unnecessary, because the assignment via `Split('/')` doesn't return `null`.

ebozduman:
I'll start reviewing this in the morning.

holterbades:
Maybe i gonna add some tests so that this error does not happen again.
Is there a special reason why in `FunctionalTest` only Streams are used as data source (`GenerateStreamFromSeed`)?
I would like to add a consistency check with an hash value or a full content check to the test case `Minio.Functional.Tests.FunctionalTest.GetObject_Test2` so that there is a validation that the upload matches the download.

ebozduman:
Sounds great @holterbades !

ebozduman:
> To the reviewer: I also noticed a minor codesmell: In line 110 and 113 there are null-checks to the `path` variable, which are unnecessary, because the assignment via `Split('/')` doesn't return `null`.

Those are not null checks. 
They are there to see if there is 1 or more than 1 elements in the `path`.
So, 1 means the command deals with a bucket, and more than 1 element in the `path` means the command deals with an object.

holterbades:
> Those are not null checks.
They are there to see if there is 1 or more than 1 elements in the path.

Sure there are null-checks; I mean the `?` in the `path?.ToList()`. If you have `var path = someUri.Split('/');` in line 104, `path` can't be null because `Split(...)` does not return `null`. The least `.Split()` can return is a `string[1] { "" }` if the input is `""`.

Also it would be more efficient to access the element count of an array via `path.Length` and not convert it to a list twice. 

I changed it accordingly.

holterbades:
I would hand in changes to the tests in a separate PullRequest from my private account some time later.

holterbades:
I don't get it why the assertion is now failing but did not fail the last executions of this testrun before the error handling bug . But in my opinion this testcase is wrong: The server responds with an error, which is expected in this test case, but the error message which is returned is checked if it is exactly the expected message. In my opinion the test case should only check if the error message starts with the expected text.

Furthermore there was another test bug which was introduced in the same commit mentioned in #1041 and was not discovered because of the removed error handling.

holterbades:
hmmmh ... The test is now failing at a new location. Locally this testcase is working just fine. Maybe a race condition in the test itself? I really don't think this has something to do with my fix.

holterbades:
@ebozduman Hey, I would need some help with that PR. I don't know what is going on with this tests.

ebozduman:
@holterbades 

There is another PR waiting for to be reviewed, PR#1043, and I think it can help resolve some of these test failure issues.
Let's wait for it to be merged, then rebase your/this PR and let's see how it goes with this PR.
You are more then welcome to check it out and review it, if you have time.

holterbades:
@ebozduman 
Since I suspect that this #1043 will take even longer, I would strongly suggest [unlisting the currently broken nuget package from nuget.org](https://learn.microsoft.com/en-us/nuget/nuget-org/policies/deleting-packages), so that no user can install this broken version.

ebozduman:
I just ran `dotnet regitlint` on your branch for the format issue(s).
I think the build test will run smoothly and all finish successfully.
Then I can merge the PR.

ebozduman (approved):
LGTM

PR #1077, opened 2024-05-17
A possible fix for random progress issue

ebozduman:
Progress' transferred data is not updated from time to time during the build functional tests and causes the transferred data/object size assertion in test is `PutObject_Test9` to fail.
This happens randomly and only during the build testing and I cannot reproduce this issue in my local environment.
That's why I created this PR to test if this fix is a correct workaround or not.
If the same problem keeps happening again, then I will revert the change back.


ebozduman:
This PR is not good.
It'll be closed.

ramondeklein:
@ebozduman I provided a fix in #1079.

PR #1101, opened 2024-06-05
Fix objects with spaces in name not returned correctly fixes #1100

### LINKED ISSUE #1100: Files uploaded with spaces in the name are returned with pluses in ListObjectsEnumAsync
similar to issue #342 the new method ListObjectsEnumAsync suffers from the problem where the item's Key contains a plus where the space was, so if the result are used for subsequent requests the item isn't found.

A `Minio.Exceptions.UnexpectedMinioException: MinIO API responded with message=The specified key does not exist.` exception is raised

heedfull:
The Uri.UnescapeDataString method has been replaced with the HttpUtility.UrlDecode method in the BucketOperations.cs file to correctly decode the item Key. Additionally, the ListObjects_Test7 function has been added to the functional tests to test whether ListObjects lists all objects matching a prefix with a space in a non-recursive manner.

### DIFF
--- a/Minio.Functional.Tests/FunctionalTest.cs
+++ b/Minio.Functional.Tests/FunctionalTest.cs
@@ -5335,6 +5335,50 @@ public static class FunctionalTest
             await TearDown(minio, bucketName).ConfigureAwait(false);
         }
     }
+    
+    internal static async Task ListObjects_Test8(IMinioClient minio)
+    {
+        var startTime = DateTime.Now;
+        var bucketName = GetRandomName(15);
+        var prefix = "minix+";
+        var objectName = prefix + GetRandomName(10);
+        var args = new Dictionary<string, string>
+            (StringComparer.Ordinal)
+            {
+                { "bucketName", bucketName },
+                { "objectName", objectName },
+                { "prefix", prefix },
+                { "recursive", "false" }
+            };
+        try
+        {
+            await Setup_Test(minio, bucketName).ConfigureAwait(false);
+            var tasks = new Task[2];
+            for (var i = 0; i < 2; i++)
+                tasks[i] = PutObject_Task(minio, bucketName, objectName + i, null, null, 0, null,
+                    rsg.GenerateStreamFromSeed(1));
+
+            await Task.WhenAll(tasks).ConfigureAwait(false);
+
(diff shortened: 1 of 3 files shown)


harshavardhana (changes requested):
We need to add support for encoding-type=url for List() and keep the URL escape.

heedfull:
I'm not really sure what you mean. The new functional test I added fails without the change submitted. This is the same fix that was used for #343 

There are no new failing tests because of these changes

heedfull:
> We need to add support for encoding-type=url for List() and keep the URL escape.

that's not what I was hoping to achieve with this PR. I want to add support for objects with spaces in their name. Maybe if you can give more details in separate issue then I can take a look.

harshavardhana:
> > We need to add support for encoding-type=url for List() and keep the URL escape.
> 
> that's not what I was hoping to achieve with this PR. I want to add support for objects with spaces in their name. Maybe if you can give more details in separate issue then I can take a look.

Try uploading an object with + in its name this PR will break.

heedfull:
I've added a test for a prefix with a plus character in it. It passes before and after the change.

harshavardhana:
The reason this works is its already added https://github.com/minio/minio-dotnet/blob/5226fa47663110f05566d24a24cd3f85216d4056/Minio/DataModel/Args/GetObjectListArgs.cs#L96

harshavardhana (dismissed):

harshavardhana (approved):

heedfull:
Anything I can do to help get this PR merged?

ebozduman:
Sorry @heedfull 
I did not notice the progress in this PR.
Running the build tests now.
As soon ass the build test is complete, I'll merge it.

heedfull:
no worries, thanks for merging.

Any plans to ship a new version of nuget package?

ebozduman:
@heedfull 

> Any plans to ship a new version of nuget package?


Yes. It is in the plan.
There is a regression problem we are working on.
A new nuget package will be generated as soon as the regression issue is resolved.

heedfull:
do you have a failing test for the regression? anything I can help with?

ebozduman:
@heedfull 

Thank you.
I filed issue #1105 for it.
The culprit is a commit in PR#1057. I have a fix and will create the PR in a bit.

Issue #1118, opened 2024-06-30
BucketExistsAsync returns true even if it does not exist


RoyalScribblz:
The logic for BucketExistsAsync is incorrect. All it determines is that the response Exception isn't a BucketNotFoundException. This means if you have another exception such as HttpRequestException, it responds that the bucket exists true even when it might not.

https://github.com/minio/minio-dotnet/blob/2e46392e2d254488b9297949863c517f465ed62f/Minio/ApiEndpoints/BucketOperations.cs#L83-L85

I was getting a "The SSL connection could not be established, see inner exception." and never knew until I stepped into the lib code.

albracko:
This is not the only method that has wrong exception handling. I just lost 4 hours trying to figure out why `BucketExistsAsync` returns true, yet method `ListBucketsAsync` returns an object that was created with default constructor. 

In my case i had wrong endpoint, but NO infor about this in logs or any exception.

I looked up your code, and the implementation of `ExecuteTaskCoreAsync` is not in-line with error handling in your API methods. All errors are basically being swallowed, leaving developers searching for a true cause why the API is not working.

RoyalScribblz:
Yes @albracko I also experienced this with putting and getting objects, I spent a while trying to figure out why I get `""` response back from getting object but turns out whole time it was never even putting anything in there in the first place as there was no connection but it swallowed exception.

rekarpcpro:
I think exception handling needs a review.

my code:

bool exists = await minioClient.BucketExistsAsync(new BucketExistsArgs().WithBucket(bucketName)); 

I will get true even if the minio container be down and not be accessible by any mean 🙄

jculverwell:
Any progress on this?

jculverwell:
Swallowing all the errors is really disappointing.  I spent a whole day investigating why the Minio dotnet Client was saying it could find a bucket I couldn't see it.  Thankfully I found this github issue.  To find the actual issue I ended up installing the AWSSDK.S3 nuget package and using the code below.  This immediately surfaced an SSL issue.  

Hope you guys can get this fixed.

`using Amazon;
using Amazon.S3;


public class TestMinioAWS
{
    private readonly AmazonS3Client _s3Client;

    public TestMinioAWS()
    {
        var endpoint = "minio.minio-tenant.svc.cluster.local";
        var accessKey = "miniouser";
        var secretKey = "xxxxxxx";

        var config = new AmazonS3Config
        {
[...]

PR #1120, opened 2024-07-01
#1118 throw exceptions returned from BucketExistsAsync

RoyalScribblz:
Throw exceptions returned in the response as BucketNotFoundException is already caught.

### DIFF
--- a/Minio/ApiEndpoints/BucketOperations.cs
+++ b/Minio/ApiEndpoints/BucketOperations.cs
@@ -80,9 +80,8 @@ public partial class MinioClient : IBucketOperations
             using var response =
                 await this.ExecuteTaskAsync(ResponseErrorHandlers, requestMessageBuilder,
                     cancellationToken: cancellationToken).ConfigureAwait(false);
-            return response is not null &&
-                   (response.Exception is null ||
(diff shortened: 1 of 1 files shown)


RoyalScribblz:
#1120

michaelwv:
Any estimated date for releasing this bugfix?

harshavardhana (changes requested):
BucketExistsAsync must be returning false and no exceptions when bucket does not exist. 

Please add a functional test to cover this scenario // cc @ebozduman

ebozduman:
Checking the code and the existing test to see what is missing to capture the thrown exception as a failure...

albracko:
Just to add here..all API methods should be checked not just this one, because Error Handling for other API methods is also wrong

COCPORN:
I also ran into this. If the AccessKey or SecretKey is wrong, checking if a bucket exists returns true.

ebozduman:
@RoyalScribblz , @michaelwv, @albracko , @COCPORN 
I appreciate it if you could test and verify the fix, #1141 (not merged yet as of 07/24 23:42), against your scenarios.
If you are satisfied with your testing, you/I can close this PR.

PR #1110, opened 2024-06-24
Add support for listening to global events

adelinag08:
This PR implements the following:

- Introduce two new types of EventType that represent global events ( s3:BucketCreated:* and s3:BucketRemoved:*)
- Add `ListenNotificationsAsync`  that uses `ListenBucketNotificationsAsync` to listen for global notifications without a bucket name set

Resolves [ #1109](https://github.com/minio/minio-dotnet/issues/1109)

PR #1130, opened 2024-07-12
Adds missing parts of the fix in PR #1110

ebozduman:
After PR #1110 is merged, the functional test case added in PR #1110 started failing sporadically but frequently.
This PR fixes the timing issue and there were bunch of unnecessary code that is removed.

### DIFF
--- a/Minio.Functional.Tests/FunctionalTest.cs
+++ b/Minio.Functional.Tests/FunctionalTest.cs
@@ -2658,7 +2658,7 @@ public static class FunctionalTest
                 () => { }
             );
 
-            await Task.Delay(75).ConfigureAwait(false);
+            await Task.Delay(150).ConfigureAwait(false);
             // Trigger the event by creating a new bucket
             _ = await CreateBucket_Tester(minio, bucketName).ConfigureAwait(false);
 
--- a/Minio.Functional.Tests/FunctionalTest.cs
+++ b/Minio.Functional.Tests/FunctionalTest.cs
@@ -1041,11 +1041,11 @@ public static class FunctionalTest
     {
         // Create a new bucket
         await minio.MakeBucketAsync(new MakeBucketArgs().WithBucket(bucketName)).ConfigureAwait(false);
+        await Task.Delay(1000).ConfigureAwait(false);
 
         // Verify the bucket exists
-        var bucketExists = await minio.BucketExistsAsync(new BucketExistsArgs().WithBucket(bucketName))
+        return await minio.BucketExistsAsync(new BucketExistsArgs().WithBucket(bucketName))
             .ConfigureAwait(false);
-        return bucketExists;
     }
 
     internal static async Task StatObject_Test1(IMinioClient minio)
@@ -2657,13 +2657,12 @@ public static class FunctionalTest
                 _ => { },
                 () => { }
             );
-
-            await Task.Delay(150).ConfigureAwait(false);
             // Trigger the event by creating a new bucket
             _ = await CreateBucket_Tester(minio, bucketName).ConfigureAwait(false);
 
             var eventDetected = false;
             for (var attempt = 0; attempt < 20; attempt++)
+            {
                 if (received.Count > 0)
                 {
                     // Check if there is any unexpected error returned
@@ -2704,6 +2703,7 @@ public static class FunctionalTest
                         break;
                     }
                 }
+            }
 
             subscription.Dispose();
             if (!eventDetected)
@@ -2773,13 +2773,13 @@ public static class FunctionalTest
                 () => { }
             );
 
-
             _ = await PutObject_Tester(minio, bucketName, objectName, null, contentType,
                 0, null, rsg.GenerateStreamFromSeed(1 * KB)).ConfigureAwait(false);
 
             // wait for notifications
             var eventDetected = false;
             for (var attempt = 0; attempt < 10; attempt++)
+            {
                 if (received.Count > 0)
                 {
                     // Check if there is any unexpected error returned
@@ -2827,8 +2827,9 @@ public static class FunctionalTest
                         break;
                     }
                 }
+            }
 
-            // subscription.Dispose();
+            subscription.Dispose();
             if (!eventDetected)
                 throw new UnexpectedMinioException("Failed to detect the expected bucket notification event.");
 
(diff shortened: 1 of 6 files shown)


Minio.Functional.Tests/FunctionalTest.cs
```diff
@@ -2804,7 +2825,7 @@ internal static async Task ListenBucketNotificationsAsync_Test1(IMinioClient min
                     }
                 }
 
-            // subscription.Dispose();
+            subscription.Dispose();
```
ramondeklein, line 2828: Why not use the `using` keyword when getting the subscription? That would automatically dispose the subscription when it goes out of scope.

Minio.Functional.Tests/FunctionalTest.cs
```diff
@@ -2651,23 +2649,50 @@ internal static async Task ListenNotificationsAsync_Test1(IMinioClient minio)
...
+                        var receivedJson = XmlStrToJsonStr(received[1].Json);
+
+                        // Cleanup the "Error" key encapsulating "receivedJson"
+                        // data. This is required to match and convert json data
+                        // "receivedJson" into class "ErrorResponse"
+                        var len = "{'Error':".Length;
+                        var trimmedFront = receivedJson[len..];
+                        var trimmedFull = trimmedFront[..^1];
+
+                        var err = JsonSerializer.Deserialize<ErrorResponse>(trimmedFull);
```
ramondeklein, line 2687: Why not parse as XML instead? This kind of "trimming" the JSON is error-prone. If I read it correctly you assume that it always starts with `{'Error':` and it always ends with `}`. JSON that would have inserted spaces would have been correct JSON too, so better use a regular JSON deserializer and fetch the `Error` object from it.

Even though "it just a unit-test", we should keep code clean.

ramondeklein (changes requested):

Minio.Functional.Tests/FunctionalTest.cs
ebozduman: Done!

Minio.Functional.Tests/FunctionalTest.cs
ebozduman: Actually it turned out to be the whole logic here was wrong and this part of the code was completely unnecessary.
So, I removed the part of the code, including the trimming, that would never be executed.
Now, it looks much cleaner.
I also modified the next test, `ListenBucketNotificationAsync_Test1`, which also had the same logic.
This is actually should have been addressed in a separate issue and  PR, but since  the work is already done, I did not remove it. (Btw, the same cleanup needs to be done for methods; `ListenBucketNotificationAsync_Test2` and `ListenBucketNotificationAsync_Test3`, but I'd address them in a separate issue with reference to this PR).

ebozduman: @ramondeklein 
Could you please check the code again?

ebozduman: Lint changes are in its way...

ebozduman: When I run `dotnet regitlint` it did not find any lint problem... looking into it

ebozduman: for some reason running `dotnet regitlint` locally did not  report any problem but there were some offending changes which were causing the build test format check to fail.
Adding the `--print-diff` option in the `minio-dotnet.yml --fail-on-diff` helped to see those missing lint changes. The last commit resolved the issue and all tests passed.

@ramondeklein 
Could you re-review the changes?

harshavardhana (approved):

ramondeklein (changes requested):

Minio.Functional.Tests/FunctionalTest.cs
```diff
@@ -1041,13 +1041,11 @@ internal static async Task<bool> CreateBucket_Tester(IMinioClient minio, string
     {
         // Create a new bucket
         await minio.MakeBucketAsync(new MakeBucketArgs().WithBucket(bucketName)).ConfigureAwait(false);
+        await Task.Delay(800).ConfigureAwait(false);
```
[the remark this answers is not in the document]
ebozduman: I run against `play.minio.io`, which is, I believe, an erasure coded single node configuration.

Similar delays were used in the original PR (#1110) and in the 2 other                  but I agree random numbers is not a guaranteed way the code will run successfully and consistently.

I am not familiar with a bucket creation delay about a specific configuration.

I've seen `await` not waiting enough until a newly created bucket becomes available after creation or subscription starts functioning. Without those waits, the test definitely fails.

I've created an issue, #1139, to address Task.Delay`s and some other cleanup issues in Notification tests. This way, we can merge the other 6 PRs waiting for this PR to be merged.

Minio.Functional.Tests/FunctionalTest.cs
```diff
@@ -1041,13 +1041,11 @@ internal static async Task<bool> CreateBucket_Tester(IMinioClient minio, string
...
 
         // Verify the bucket exists
-        var bucketExists = await minio.BucketExistsAsync(new BucketExistsArgs().WithBucket(bucketName))
+        return await minio.BucketExistsAsync(new BucketExistsArgs().WithBucket(bucketName))
             .ConfigureAwait(false);
-        Assert.IsTrue(bucketExists, $"Bucket {bucketName} was not created successfully.");
-
-        return bucketExists;
```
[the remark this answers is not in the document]
ebozduman: Done!

Minio.Functional.Tests/FunctionalTest.cs
```diff
@@ -2648,61 +2646,54 @@ internal static async Task ListenNotificationsAsync_Test1(IMinioClient minio)
...
+            var listenArgs = new ListenBucketNotificationsArgs()
+                .WithEvents(eventsList);
+            var events = minio.ListenNotifications(listenArgs);
             var eventDetected = false;
+            using (var subscription = events.Subscribe(
+                       received.Add,
+                       _ => { },
+                       () => { }))
+            {
+                await Task.Delay(200).ConfigureAwait(false);
```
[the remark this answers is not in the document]
ramondeklein: This is probably already an issue in the existing SDK, so maybe create an issue for this, so we can merge this PR.

ramondeklein (approved):

PR #1191, opened 2024-10-11
V5.0.0 patch release for a customer

ebozduman:
A customer complained about PutObjectAsync api not supporting zero sized objects.
They use version 5.0.0, so the patch fix is created on version 5.0.0.


ebozduman:
Did something wrong. It looks like this PR is messed up.
Closing...

PR #1193, opened 2024-10-14
V5.0.0 patch release for a customer

ebozduman:
A customer complained about PutObjectAsync api not supporting zero sized objects.
The customer uses version 5.0.0, so the patch fix is created on version 5.0.0.

### DIFF
--- a/Minio/DataModel/ObjectOperationsArgs.cs
+++ b/Minio/DataModel/ObjectOperationsArgs.cs
@@ -1838,7 +1838,7 @@ public class PutObjectArgs : ObjectWriteArgs<PutObjectArgs>
 
         if (!string.IsNullOrWhiteSpace(FileName)) Utils.ValidateFile(FileName);
         // Check object size when using stream data
-        if (ObjectStreamData is not null && ObjectSize == 0)
+        if (ObjectStreamData is not null && ObjectSize == -1)
             ObjectSize = ObjectStreamData.Length;
         Populate();
     }
--- a/Minio/DataModel/ObjectOperationsArgs.cs
+++ b/Minio/DataModel/ObjectOperationsArgs.cs
@@ -1839,7 +1839,7 @@ public class PutObjectArgs : ObjectWriteArgs<PutObjectArgs>
         if (!string.IsNullOrWhiteSpace(FileName)) Utils.ValidateFile(FileName);
         // Check object size when using stream data
         if (ObjectStreamData is not null && ObjectSize == 0)
-            throw new ArgumentException($"{nameof(ObjectSize)} must be set");
+            ObjectSize = ObjectStreamData.Length;
         Populate();
     }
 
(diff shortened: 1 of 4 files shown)


Minio/MinioClient.cs
```diff
@@ -411,10 +411,7 @@ private async Task<ResponseResult> ExecuteTaskCoreAsync(
     {
         var startTime = DateTime.Now;
         // Logs full url when HTTPtracing is enabled.
-        if (trace)
-        {
-            var fullUrl = requestMessageBuilder.RequestUri;
-        }
+        if (trace) _ = requestMessageBuilder.RequestUri;
```
ramondeklein, line 414: Why not remove this all together? What's the point of doing this?

Minio/DataModel/ObjectOperationsArgs.cs
```diff
@@ -1839,7 +1839,7 @@ internal override void Validate()
         if (!string.IsNullOrWhiteSpace(FileName)) Utils.ValidateFile(FileName);
         // Check object size when using stream data
         if (ObjectStreamData is not null && ObjectSize == 0)
-            throw new ArgumentException($"{nameof(ObjectSize)} must be set");
+            ObjectSize = ObjectStreamData.Length;
```
ramondeklein, line 1842: I don't think this is what the client asked for. The client asked that if he specified `ObjectSize = 0` that it should create an empty file. Now it won't use the `ObjectSize` parameter, but it will auto-detect the size from the stream (note that this is only possible for streams that have their `CanSeek` property set).

I think we should initialize `ObjectSize` to `-1` in the constructor and that would mean that the size will be auto-detected. When set to 0 or larger, then it should use that size instead. We should also ensure that we don't read more than is specified. So when `ObjectSize = 10` and the stream delivers 20 bytes, then we should only upload 10 bytes.

ramondeklein (changes requested):

ramondeklein:
Are we planning to deploy a new v5 SDK version?

Minio/DataModel/ObjectOperationsArgs.cs
ebozduman: Makes sense.
I was thinking more towards like `ObjectSize` being useless as it is implemented right now and maybe replacing it with auto-detect object size could be a better idea since retrieving part of a stream/file may not be always possible.

But what you've suggested makes much more sense.
On the other side, the code was like this quite for some time.

To make it easier to generate a 5.0.0 patch release package available for the customer as soon as possible, I suggest to create a separate issue for it and address it after the patch release.
Please share your opinion.

ebozduman:
Yes.
The decision is to create a patch release for the customer and the only requirement the customer asked for is to support zero-sized files/objects.

ramondeklein:
@ebozduman If they can do non-zero size files using this method, then I'm fine with that.

Minio/MinioClient.cs
ebozduman: ok. Done!

ramondeklein (approved):

harshavardhana (approved):

PR #1197, opened 2024-10-19
Bump up to 5.0.1 for release

donatello:


donatello:
I think we need to do this before release scripts will work

PR #1222, opened 2024-12-19
Fix build

ramondeklein:
Fixed the build by fixing linter issues. I have disabled the "incomplete upload" tests, because they fail often. The modified code doesn't apply to these tests.


ramondeklein:
@ebozduman If you have bandwidth, could you check why these tests started failing?

ebozduman:
@ramondeklein 
I'll look into those failures.

Issue #1068, opened 2024-05-01
Failing to upload smaller streams having .WithObjectSize() as max possible stream size


w7rus:
I'm having an ASP.NET Controller incoming form data and request size is limited with following attributes:

```csharp
//Assume Consts.AvatarMaxFileSize is 10MB

[RequestSizeLimit(Consts.AvatarMaxFileSize + 4096)]
[RequestFormLimits(MultipartBodyLengthLimit = Consts.AvatarMaxFileSize)]
...
```

Then the minioClient instance forces me to use .WithObjectSize(), assuming limits set on the controller, stream size would be less or equal to Consts.AvatarMaxFileSize, so:

```csharp
...

//Upload file
var putObjectArgs = new PutObjectArgs()
    .WithBucket(bucketName)
    .WithStreamData(file)
    .WithObject(fileName)
    .WithContentType(fileContentType);
    .WithObjectSize(Consts.AvatarMaxFileSize);

await _minioClient.PutObjectAsync(putObjectArgs, cancellationToken).ConfigureAwait(false);
```

Which throws an exception when validating CompleteMultipartUploadArgs since ETags dictionary is null, because of following block of code at ``Minio.MinioClient.PutObjectPartAsync``

```csharp
// This shouldn't happen where stream size is known.
if (partCount != numPartsUploaded && args.ObjectSize != -1)
{
    var removeUploadArgs = new RemoveUploadArgs()
        .WithBucket(args.BucketName)
        .WithObject(args.ObjectName)
        .WithUploadId(args.UploadId);
    await RemoveUploadAsync(removeUploadArgs, cancellationToken).ConfigureAwait(false);
    return null;
}

//args.ObjectSize = 52428800 - stream max possible size
//partSize = 16777216 - stream actual size
//partCount = 4
//numPartsUploaded = 1
```

This happens when stream size is smaller than maximum possible, causing partCount != numPartsUploaded to return null ETags, removing an object.

And another excessive check to remove

```
if (bytesRead != (int)args.ObjectSize)
    throw new UnexpectedShortReadException($"Data read {bytesRead.ToString(CultureInfo.InvariantCulture)} is shorter than the size {args.ObjectSize.ToString(CultureInfo.InvariantCulture)} of input buffer.");
```

I can not determine actual file size since i read multipart request as stream, dropping it in memory/temp file does not sound like a good solution, rather than streaming directly

w7rus:
> +1
> 
> Did you find solution?

[...]

PR #1235, opened 2025-01-05
Allow uploading smaller streams than set via .WithObjectSize()

w7rus:
When working with large files, you may not want to drop them in the disk in application logic to determine actual file length before upload, but rather stream it directly to Minio.

With changes made, .WithObjectSize() acts a max cap the stream can be, and smaller streams are also accepted.


w7rus:
Related issue: https://github.com/minio/minio-dotnet/issues/1068

klauspost (changes requested):
This can cause silent corruption for users. 

I would be supporting a change that doesn't modify existing behavior and for example adds a `.WithExpectedObjectSize()`, which forces multipart upload, with part sizes set for the approximate size given, but will not fail on a different size.

w7rus:
While making changes, found out you can set .WithObjectSize(-1), and it will work as needed.

I'll close related issue. It would be good to have documentation on that.

PR #1238, opened 2025-01-07
stops Headers overrite the user defined ContentType

### LINKED ISSUE #1118: BucketExistsAsync returns true even if it does not exist

ebozduman:
Fixes #1118

### DIFF
--- a/Minio.Functional.Tests/FunctionalTest.cs
+++ b/Minio.Functional.Tests/FunctionalTest.cs
@@ -1014,8 +1014,8 @@ public static class FunctionalTest
                 .WithStreamData(filestream)
                 .WithObjectSize(size)
                 .WithProgress(progress)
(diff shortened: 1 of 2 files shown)


ebozduman:
Tested and verified the fix.
Modified the existing test, `PutObject_Test1` to cover this scenario.

harshavardhana (approved):

PR #1320, opened 2025-07-11
adds the missing invalid access&secret key exceptions

ebozduman:
Before this fix, what we caught and displayed for an invalid access key value was:
```
... "Alert":"Object reference not set to an instance of an object.",
"Message":"System.NullReferenceException: Object reference not set to an instance of an object. ...
```

while the stack trace showed (this is the correct exception message).
```
Unhandled exception. Minio.Exceptions.AuthorizationException: The Access Key Id you provided does not
exist in our records. ...
```

For an invalid secret key, what we catch is:
```
"Alert":"Object reference not set to an instance of an object."
```

and the stack trace is:
```
Unhandled exception. Minio.Exceptions.AuthorizationException: The request signature we calculated does
not match the signature you provided. Check your key and signing method.
```

It looks like the stack trace is more accurate than vague Minio exceptions.

**NEW  BEHAVIOR**

First of all, the thrown exceptions should be the same in Minio client and the system Stacktrace output.

Here is the new behavior.

**With invalid Endpoint:**
```
Minio client:

System.Net.Http.HttpRequestException: Name or service not known (play.min.ioxxx:443)
 ---> System.Net.Sockets.SocketException (0xFFFDFFFF): Name or service not known
```
```
Stacktrace:

Unhandled exception.System.Net.Http.HttpRequestException: Name or service not known (play.min.ioxxx:443)
 ---> System.Net.Sockets.SocketException (0xFFFDFFFF): Name or service not known
```

**With invalid AccessKey:**
```
Minio client:

MinIO API responded with message=Access denied on the resource: 
/<bucket-name>/","Message":"Minio.Exceptions.AccessDeniedException
```
```
Stacktrace:

Unhandled exception. Minio.Exceptions.AccessDeniedException: 
MinIO API responded with message=Access denied on the resource: /<bucket-name>/
```

**With invalid SecretKey:**
```
Minio client:

MinIO API responded with message=Access denied on the resource: 
/<bucket-name>/","Message":"Minio.Exceptions.AccessDeniedException
```
```
Stacktrace:

Unhandled exception. Minio.Exceptions.AccessDeniedException: 
MinIO API responded with message=Access denied on the resource: /<bucket-name>/
```

### DIFF
--- a/Minio.Examples/Program.cs
+++ b/Minio.Examples/Program.cs
@@ -135,7 +135,7 @@ public static class Program
             else
                 Console.WriteLine();
         });
-        var objectsList = [];
+        List<string> objectsList = [];
         for (var i = 0; i < 10; i++)
             objectsList.Add(objectName + i);
 
--- a/Minio/DataModel/Args/RemoveBucketArgs.cs
+++ b/Minio/DataModel/Args/RemoveBucketArgs.cs
@@ -1,4 +1,4 @@
-﻿/*
+/*
  * MinIO .NET Library for Amazon S3 Compatible Cloud Storage, (C) 2020, 2021 MinIO, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");


AmBplus:
Check if your locale time is US or not. It seems there is a problem with datetime that may cause an unknown error. I have sent an issue report for this matter. Please verify if the situation is the same or not.
https://github.com/minio/minio-dotnet/issues/1323

PR #1330, opened 2025-09-12
Fix condition on nullable property

Metalaka:
`ResponseResult.Exception` is nullable by usages. We have to check that nullability for standard Error cas.
eg when the Minio client SecretKey is invalid, the error is handled by this method and no exception is present in the ResponseResult object.

This fix the null point exception that is thrown at this line and allow to properly retrieve the error message.


ebozduman:
@Metalaka 
#1320 is merged.
Could you please check if the issue is still reproducable?
If not reproducble, please close both the PR and the issue.

ebozduman:
@Metalaka 
Closing as this PR as it became obsolete.
PR #1320 removed the part of the code this PR made the change..

PR #1340, opened 2025-10-24
Fix AddMinio registration lifetime

### LINKED ISSUE #1268: MinioClient Injection Logic with ServiceLifetime Can Be Confusing
I’ve been reviewing the code for dependency injection of MinioClient in the minio-dotnet SDK, and I noticed some potential confusion in the way the client is being injected based on the ServiceLifetime provided.

In the code:
https://github.com/minio/minio-dotnet/blob/master/Minio/ServiceCollectionExtensions.cs
```
var client = minioClientFactory.CreateClient();
client.Config.ServiceProvider = services.BuildServiceProvider();
switch (lifetime)
{
    case ServiceLifetime.Singleton:
	services.TryAddSingleton(_ => client);
	break;
    case ServiceLifetime.Scoped:
	services.TryAddScoped(_ => client);
	break;
    case ServiceLifetime.Transient:
	services.TryAddTransient(_ => client);
	break;
}
```
The issue is that the client is being created first and then injected as a singleton, scoped, or transient instance based on the lifetime provided. However, it seems like the code is always injecting the same instance of MinioClient, which makes the ServiceLifetime setting effectively irrelevant.

whether it's recommended to use MinioClient as a Singleton or as a Transient/Scoped instance in a typical application?

Dreamescaper:
Fixes #1268 

This PR fixes the following:
- Lifetime parameter now actually works (it was always singleton effectively before)
- It doesn't build IServiceProvider from IServiceCollection unnessesseraly
- It doesn't create IMinioClientFactory or IMinioClient until requested

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -16,7 +16,7 @@
         <RepositoryUrl>https://github.com/minio/minio-dotnet</RepositoryUrl>
         <RepositoryType>git</RepositoryType>
         <Product>$(AssemblyName) ($(TargetFramework))</Product>
-        <Version>7.0.0</Version>
+        <Version>7.0.1</Version>
 
         <!--<Nullable>enable</Nullable>-->
         <ImplicitUsings>enable</ImplicitUsings>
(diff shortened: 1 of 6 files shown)


ebozduman:
@Dreamescaper 
There is a build error reported.
I checked quickly and it is `error IDE0032: Use auto property` for lines 26 and 27 in `ResponseResult.cs` file.
PTAL.
You may want to suppress IDE0032 for these 2 lines.

PR #1366, opened 2026-03-02
Publish symbols in NuGet package and add license

ramondeklein:
Update NuGet package generation:
- Also publish source-code with the package to make debugging easier.
- Set license to Apache 2.0 in NuGet specification.


shtripat (approved):
lgtm

PR #1368, opened 2026-03-04
Add integration tests for bucket operations

shtripat:
Bonus: fix Content-MD5 and lifecycle XML deserialization

### Summary

- Add integration tests for bucket encryption, lifecycle, replication, and policy operations
- Fix missing Content-MD5 header on SetBucketEncryptionAsync, SetBucketLifecycleAsync, SetBucketReplicationAsync, and DeleteObjects
- Fix LifecycleConfiguration deserialization to handle XML responses without an S3 namespace declaration

#### Integration tests (BucketTests.cs)

Four new tests covering the previously untested bucket operation groups:
- BucketEncryptionWithoutConfig — verifies GetBucketEncryptionAsync throws ServerSideEncryptionConfigurationNotFoundError when no config exists. A full Set/Get/Remove round-trip is not covered because the standard MinIO container requires KMS for bucket-default SSE.
- BucketLifecycle — full round-trip: null on fresh bucket → set a rule → verify it round-trips → remove → back to null
- BucketReplicationWithoutConfig — verifies GetBucketReplicationAsync returns null on a bucket with no replication config
- BucketPolicy — full round-trip: null on fresh bucket → set a read-only policy → verify JSON structure → remove → back to null

#### Bug fix: missing Content-MD5 header (MinioClient.cs)

MinIO requires a Content-MD5 or x-amz-checksum-* header on certain PUT requests. Without it the server returns 400 BadRequest: Missing or invalid required header. Added a private AddContentMd5Async helper and applied it to the three Set* bucket configuration operations. Also used it to replace the existing inline TODO in DeleteObjects.

Bug fix: namespace-agnostic lifecycle deserialization (LifecycleConfiguration.cs)

MinIO's Go server marshals lifecycle XML without emitting a namespace declaration (<LifecycleConfiguration> rather than <LifecycleConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/">). The deserialization methods were looking up elements using the S3 namespace prefix and found nothing, returning an empty rule list. Fixed by detecting the namespace from the root element via GetDefaultNamespace() (which returns XNamespace.None when absent) and passing it through to DeserializeRule and DeserializeTag. Serialization is unchanged and continues to emit the correct S3 namespace.


ramondeklein:
@shtripat Great catch, but this is a broader issue:

> The XML namespace historically associated with Amazon S3 is `http://s3.amazonaws.com/doc/2006-03-01/`. This URI appears in many S3-style XML documents, especially older ones, for example:
> ```xml
> <ListBucketResult xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
>   …
> </ListBucketResult>
> ```
> Even though this namespace exists, it is not part of the protocol contract. AWS does not require clients to validate or enforce it, and many official S3 responses omit it entirely. The namespace is effectively a decorative identifier rather than a functional one.

We shouldn't add it to requests and remove it from all responses. I'll update the SDK in this PR.

ramondeklein:
@shtripat PTAL...

Minio.IntegrationTests/Tests/BucketTests.cs
```diff
@@ -116,4 +117,113 @@ public async Task BucketObjectLock()
         Assert.Equal(RetentionMode.Compliance, ruleYears.Mode);
         Assert.Equal(3, ruleYears.Years);
     }
+
+    [Fact]
+    public async Task BucketEncryptionWithoutConfig()
```
ramondeklein, line 127: We may want to add a test that does set bucket encryption too.

Minio.IntegrationTests/Tests/BucketTests.cs
```diff
@@ -116,4 +117,113 @@ public async Task BucketObjectLock()
...
+        var rc = await client.GetBucketReplicationAsync("test").ConfigureAwait(true);
+        Assert.Null(rc);
+    }
+
+
+    [Fact]
+    public async Task BucketPolicy()
+    {
+        var client = CreateClient();
+        await client.CreateBucketAsync("test").ConfigureAwait(true);
```
ramondeklein: You may want to use `const string bucketName = "test"` if the bucket name is repeated often.

ramondeklein (approved):
Some minor remarks.

Minio/Implementation/MinioClient.cs
```diff
@@ -1620,6 +1615,16 @@ private HttpRequestMessage CreateRequest(HttpMethod method, string path, XElemen
         return req;
     }
 
+    private static async Task AddContentMd5Async(HttpRequestMessage req, CancellationToken cancellationToken)
```
ramondeklein, line 1701: I think we should check which API requests require this to be set.

ramondeklein (changes requested):

Minio/Implementation/MinioClient.cs
ramondeklein: Copilot suggest this, but we need to confirm:
image

shtripat: will check this out and do the needful

shtripat: Here is detailed expectation
```
  Required by the S3 spec — currently missing AddContentMd5Async:                                                                                                                     
                                                                                                                                                                                      
  ┌─────────────────────────────────┬───────────────────────┬───────┐                                                                                                                 
  │             Method              │       Operation       │ Line  │                                                                                                                 
  ├─────────────────────────────────┼───────────────────────┼───────┤                                                                                                                 
  │ SetBucketTaggingAsync           │ PUT ?tagging          │ ~198  │                                                                                                                 
  ├─────────────────────────────────┼───────────────────────┼───────┤                                                                                                                 
  │ SetObjectLockConfigurationAsync │ PUT ?object-lock      │ ~841  │                                                                                                                 
  ├─────────────────────────────────┼───────────────────────┼───────┤                                                                                                                 
  │ SetObjectLegalHoldAsync         │ PUT ?legal-hold       │ ~1083 │
  ├─────────────────────────────────┼───────────────────────┼───────┤
  │ SetObjectRetentionAsync         │ PUT ?retention        │ ~1113 │
  ├─────────────────────────────────┼───────────────────────┼───────┤
  │ ClearObjectRetentionAsync       │ PUT ?retention        │ ~1128 │
  ├─────────────────────────────────┼───────────────────────┼───────┤
  │ SetObjectTagsAsync              │ PUT ?tagging (object) │ ~1419 │
  └─────────────────────────────────┴───────────────────────┴───────┘

  Required — correctly implemented:

  ┌───────────────────────────┬──────────────────┐
  │          Method           │    Operation     │
  ├───────────────────────────┼──────────────────┤
  │ DeleteObjectsAsync        │ POST ?delete     │
  ├───────────────────────────┼──────────────────┤
  │ SetBucketLifecycleAsync   │ PUT ?lifecycle   │
  ├───────────────────────────┼──────────────────┤
  │ SetBucketReplicationAsync │ PUT ?replication │
  └───────────────────────────┴──────────────────┘

  Has it but not strictly required (optional per spec — no harm):

  ┌──────────────────────────┬─────────────────┐
  │          Method          │    Operation    │
  ├──────────────────────────┼─────────────────┤
  │ SetBucketEncryptionAsync │ PUT ?encryption │
  └──────────────────────────┴─────────────────┘

  No body or Content-MD5 optional — correctly omitted:

  - CreateBucketAsync, SetBucketVersioningAsync, SetBucketNotificationsAsync, SetPolicyAsync, CompleteMultipartUploadAsync, CopyObjectAsync

```

ramondeklein (approved):

ramondeklein:
@shtripat I have added the commits to remove all the namespace stuff. PTAL...

